PR #2, opened 2022-07-15
Add a module and role for testing purposes

mkanoor:
Added a module called upcase that upper cases a string
Added a role call test-role that calls the module upcase


mkanoor:
@benthomasson Please review

PR #6, opened 2022-08-03
Added json schema for range plugin

mkanoor:
The json schemas are in separate files one for input and one
for output. The input could most probably be used by the UI
the output is what would be sent to the drools engine as object
definition


mkanoor:
@benthomasson Please review

bzwei:
@mkanoor why the schemas are not placed in the schema folder?

mkanoor:
@bzwei this is just a trial balloon we can discuss where it should sit. The schema folder I thought is for the ruleset and not for individual plugins. I kept it closer to where the source plugin is defined.

benthomasson (approved):

PR #8, opened 2022-08-05
add github workflow to run sanity tests

Alex-Izquierdo:
Depends on https://github.com/benthomasson/benthomasson-eda/pull/7

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,40 @@
+name: Run ansible tests
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
Notice that there is a specific action available to run ansible-test: https://github.com/marketplace/actions/ansible-test
But after several tries, I couldn't make it working because this repo doesn't follow the expected path for the collection. 

I will try in the future more tests to figure out the right config for this action (or report a bug in that project) but for now, the current approach covers our needs.

PR #10, opened 2022-08-11
Add ansible-test unit test for kafka plugin

bzwei:
passed test command
```
ansible-test units --docker default
```


bzwei:
@benthomasson @hsong-rh please review

benthomasson/eda/tests/unit/requirements.txt
```diff
@@ -0,0 +1,4 @@
+asyncio
```
Alex-Izquierdo, line 1: I think is better to have a `test_requirements.txt` file in `/tests` folder because are needed as well for integration tests. 
Ansible-galaxy doesn't take care of this, I had to follow the same approach for collection dependencies here: https://github.com/benthomasson/benthomasson-eda/pull/9

Alex-Izquierdo (changes requested):

benthomasson/eda/tests/unit/event_source/test_kafka.py
```diff
@@ -0,0 +1,45 @@
...
+            raise StopAsyncIteration
+
+
+class MockConsumer(AsyncMock):
+    def __aiter__(self):
+        return AsyncIterator()
+
+
+@pytest.mark.asyncio
+def test_reeive_from_kafka_place_in_queue(myqueue):
```
mkanoor, line 41: @bzwei Is there a typo here, should it be receive instead of reeive?

mkanoor:
@bzwei How would we ask the plugin to stop, does it handle the async task cancel. Please check with @hsong-rh we had to implement something similar for WebHook plugin to get a clean shutdown from the plugins.

bzwei:
@mkanoor This can be done in the rules. In my kafka sample rules https://github.com/benthomasson/benthomasson-eda/blob/main/benthomasson/eda/rules/kafka-test-rules.yml#L17 it calls a shutdown action corresponding to.a stop event.

mkanoor:
@bzwei this is outside of the shutdown, if the asyncio task is canceled we have to catch the exception and exit the plugin.

We need to catch this exception and exit cleanly

```
except asyncio.CancelledError:
         print("Plugin Task Cancelled")
```

There is an implementation of this and a test for it in the web hook plugin #11

bzwei:
@mkanoor The kafka plugin has a finally block where the consumer is properly stopped. https://github.com/benthomasson/benthomasson-eda/blob/main/benthomasson/eda/plugins/event_source/kafka.py#L50

Unlike the webhook unit test where we can set up a webserver and actually POST a request, for kafka plugin uni test we don't have an actual kafka server. The unit test mainly tests the logic to parse a kafka event and place it in our internal queue. I created a mocked kafka consumer that does not connect to long living kafka server.

benthomasson (approved):

PR #14, opened 2022-08-26
Add webhook integration tests

ttuffin:
Includes kafka integration test from PR12 since both tests share the same utils.


ttuffin:
@Alex-Izquierdo please review.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
```diff
@@ -0,0 +1,81 @@
...
+        requests.post(url, data=msg, headers=headers)
+
+    result = cli_runner.stderr.read().decode()
+
+    assert "'msg': 'SUCCESS'" in result
+    assert headers["user-agent"] in result
+    assert f"'Host': '127.0.0.1:{port}'" in result
+
+    cli_runner.poll()
+    assert cli_runner.returncode == 0
```
Alex-Izquierdo, line 63: You are not handling when the process has terminated. 
I think the right way is to use communicate() with timeout
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
```diff
@@ -0,0 +1,81 @@
...
+    headers = {"user-agent": "webhook_client/1.0.1"}
+
+    os.environ["WH_PORT"] = str(port)
+
+    cli_runner = factory_cli_runner(env_vars="WH_PORT")
+
+    for msg in msgs:
+        requests.post(url, data=msg, headers=headers)
+
+    result = cli_runner.stderr.read().decode()
```
Alex-Izquierdo, line 56: this should be the stdout, not stderr, otherwise it would be a bug, isn't?

benthomasson/eda/tests/integration/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    def run(self):
+        args = self._process_args()
+        return subprocess.run(
+            args, cwd=self.cwd, capture_output=True, timeout=self.timeout, check=True
+        )
+
+    def run_in_background(self):
+        args = self._process_args()
+        proc = subprocess.Popen(
+            args, cwd=self.cwd, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE
```
Alex-Izquierdo, line 62: You should not discard the stdout, why are you doing it?

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
```diff
@@ -0,0 +1,81 @@
...
+
+    assert "'msg': 'SUCCESS'" in result
+    assert headers["user-agent"] in result
+    assert f"'Host': '127.0.0.1:{port}'" in result
+
+    cli_runner.poll()
+    assert cli_runner.returncode == 0
+
+
+def test_webhook_source_error_handling(factory_cli_runner):
```
Alex-Izquierdo, line 66: If this test only check the case when the port it's already used, named it accordingly. For example: `test_webhook_source_with_busy_port`

Alex-Izquierdo (changes requested):
Don't include code from other PR's, instead mention a dependency.

benthomasson/eda/tests/integration/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        # Ensure ansible-events is running
+        for i in range(10):
```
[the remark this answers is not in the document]
ttuffin: We require this logic to ensure ansible-events starts up and is "ready", otherwise we hit a race condition where the client tries to post data before the webhook server is ready. If we move this logic to the caller then we would just repeat the logic in every other source plugin that will use this method, which seems pointless imo.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
```diff
@@ -0,0 +1,81 @@
...
+
+        runner = CLIRunner(
+            rules=ruleset, verbose=True, envvars=env_vars
+        ).run_in_background()
+
+        runners.append(runner)
+        return runner
+
+    yield _cli_runner
+    runners[0].terminate()
```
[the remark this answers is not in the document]
ttuffin: > What is the sense to use a list if you will call later only the first item?

Yeah I am not a big fan of that. I tried instead to use a nonlocal variable but that did not work, so open to suggestions for a better implementation.

> Anyway, I don't think that this fixture is justified. It makes no sense to end a process that is expected to end. If the process ends or not or when does it, depends on the test.

We could certainly achieve the same without the fixture. My rational was that a fixture with teardown would be useful for future tests that do not post a "shutdown" to the webhook server and as a failsafe to ensure the process is terminated. If we plan for every test to post a shutdown, or send it's own terminate command then teardown is unnecessary. I'm happy to go either way.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
ttuffin: No, all ansible-events output goes to stderr. There's a discussion ongoing regarding this, but for now the test needs to check stderr.

Alex-Izquierdo: Nope, not all goes to stderr as you can see in the kafka test: https://github.com/benthomasson/benthomasson-eda/pull/12/files#diff-fb34777d22053fb6ad6a8c60fb187d7dd087a8bdcab8443b7eccc7d5d3ffbb8eR36

Any standard output should go to stdout so seems this a bug. We can check stderr for now but please, add a TODO comment pointing to the respective bug.

benthomasson/eda/tests/integration/utils.py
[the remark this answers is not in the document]
Alex-Izquierdo: What if you run a process that you expect to fail? If you don't want to repeat code you can encapsulate it, but that responsibility clearly does not belong to this method.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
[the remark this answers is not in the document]
Alex-Izquierdo: Anyway you are still having the list issue, you must iterate over it.

benthomasson/eda/tests/integration/utils.py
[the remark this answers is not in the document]
ttuffin: I have moved the logic to the caller as suggested.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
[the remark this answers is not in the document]
ttuffin: Iterator now in place.

benthomasson/eda/tests/integration/event_source_webhook/test_webhook_source.py
ttuffin: TODO comment added.

Alex-Izquierdo (changes requested):
still has code from other PR's. 
https://github.com/benthomasson/benthomasson-eda/pull/14#pullrequestreview-1086823493

benthomasson:
@ttuffin please rebase.

Alex-Izquierdo (approved):

PR #16, opened 2022-09-01
Ensure arg 'delay' is always an integer

ttuffin:
Typing is not enforced for "delay" and if, for example, you set it to "5" instead of 5 then ansible-events will not loop; it runs once and exits because asyncio.sleep will not accept a string. This is a problem if you use jinja templating in your rulebook since it will always parse as a string. 

https://issues.redhat.com/browse/AAP-5497


ttuffin:
@Alex-Izquierdo @hsong-rh @benthomasson pls review.

Alex-Izquierdo (approved):

benthomasson (approved):

PR #29, opened 2022-10-17
[AAP-6385] fix ansible dependencies for install-rulebook role

Alex-Izquierdo:
Fixes AAP-6385
Ansible and ansible-runner are indirect dependencies and must be managed in the system regardless of ansible-rulebook. The playbook started to install them by convenience. But since can be installed in a system scope, the playbook can fail trying to update them, so to avoid errors, better to move as "present" if those packages were previously installed by the user.

### DIFF
--- a/roles/install_ansible_rulebook/tasks/install_fedora.yml
+++ b/roles/install_ansible_rulebook/tasks/install_fedora.yml
@@ -32,15 +32,15 @@
   ansible.builtin.pip:
     name:
       - pip
+      - wheel
     state: latest
 
-- name: Install pip dependencies
+- name: Ensure ansible dependencies
   ansible.builtin.pip:
     name:
       - ansible
       - ansible-runner
-      - wheel
-    state: latest
+    state: present
 
 - name: install ansible-rulebook
   environment:
--- a/roles/install_ansible_rulebook/tasks/install_macosx.yml
+++ b/roles/install_ansible_rulebook/tasks/install_macosx.yml
@@ -37,15 +37,15 @@
   ansible.builtin.pip:
     name:
       - pip
+      - wheel
     state: latest
 
-- name: Install pip dependencies
+- name: Ensure ansible dependencies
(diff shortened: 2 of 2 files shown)


ttuffin (approved):

.github/workflows/integration-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.9"
```
konstruktoid, line 19: `python-version: "3.8"` is still okay? 
https://github.com/ansible/ansible-rulebook/commit/79892c2d1c168de6f9ae1842fcb940606086a0b4

Alex-Izquierdo: yeah, unmade. thanks @konstruktoid

roles/install_ansible_rulebook/tasks/install_fedora.yml
```diff
@@ -32,15 +32,15 @@
...
+- name: Ensure ansible dependencies
   ansible.builtin.pip:
     name:
       - ansible
       - ansible-runner
-      - wheel
-    state: latest
+    state: present
 
 - name: install ansible-rulebook
```
ttuffin, line 45:
```suggestion
- name: Install ansible-rulebook
```

ttuffin: Can you capitalize the name on L53 as well?

Alex-Izquierdo: Can you create a PR to fix all of those issues?

ttuffin (dismissed):

bzwei (dismissed):

ttuffin (approved):

ddonahue007 (approved):
lgtm

PR #30, opened 2022-10-17
Update README with pattern description

josephpisciotta:


Alex-Izquierdo:
I am sorry, but I disagree with the approach of this document. Anyway I'm not sure if this PR is a WIP, because it seems that the last part is missing links.

TL;DR: I think it's a good text, but probably not in the right place. The installation instructions must be present.

A readme is the entry door to a project. Usually contains or refers to the guidelines to install, use, configure and contribute to the code. This seems more like the content of a commercial landing page. Not even contains a link to the document that really explains what is and how to use this code. This is a common pattern beyond the ansible community, but you can see it in other projects of the community:
https://github.com/ansible/awx/blob/devel/README.md
https://github.com/ansible/ansible/blob/devel/README.rst

I feel that the `getting started` section makes more sense in ansible-rulebook docs rather than here.

I suggest a section "about" with a brief description and a link for further details pointing to some page or document with this content. I believe that the current content of the `COLLECTION.md` is quite appropriate to the `README.md`, at least, should include a link pointing to that.

Ref: https://github.com/ansible-collections/community.general

Take into account that the README.md is automatically shown in the ansible galaxy page: https://galaxy.ansible.com/ansible/eda
And the users of the collection should be able to see there that the dependencies must be installed.

Alex-Izquierdo:
I am a bit lost. This is already merged. It was a mistake?

PR #42, opened 2022-10-28
Make alertmanager source configurable

bzwei:
Add more options to configure how to get alerts and hosts from the original data. Also allow to skip the original data and only queue the parsed alerts.

### DIFF
--- a/plugins/event_source/alertmanager.py
+++ b/plugins/event_source/alertmanager.py
@@ -12,6 +12,7 @@ Arguments:
                       Use empty string "" to treat the whole payload data as
                       one alert.
     data_host_path: The json path inside the alert data to find alerting host.
+                    Use empty string "" if there is no need to find host.
                     Default to  "labels.instance".
     data_path_separator: The separator to interpret data_host_path and
                          data_alerts_path. Default to "."
@@ -32,10 +33,9 @@ Example:
 """
 
 import asyncio
-from typing import Any, Dict
-
 from aiohttp import web
-import dpath
+from dpath import util
+from typing import Any, Dict
 
 routes = web.RouteTableDef()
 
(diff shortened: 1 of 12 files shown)


plugins/event_source/alertmanager.py
```diff
@@ -74,6 +109,12 @@ def clean_host(host):
 async def main(queue: asyncio.Queue, args: Dict[str, Any]):
     app = web.Application()
     app["queue"] = queue
+    app["data_host_path"] = args.get("data_host_path") or "labels.instance"
```
hsong-rh, line 112: Why not use `args.get("data_host_path", "labels.instance")`?

plugins/event_source/alertmanager.py
```diff
@@ -74,6 +109,12 @@ def clean_host(host):
...
+    app["data_path_separator"] = args.get("data_path_separator") or "."
```
hsong-rh, line 113: Same here: `args.get("data_path_separator", ".")`

plugins/event_source/alertmanager.py
```diff
@@ -74,6 +109,12 @@ def clean_host(host):
...
+    app["data_alerts_path"] = args.get("data_alerts_path")
```
hsong-rh, line 114: And here: `args.get("data_alerts_path", "alerts")`

plugins/event_source/alertmanager.py
```diff
@@ -74,6 +109,12 @@ def clean_host(host):
...
+    if app["data_alerts_path"] is None:
+        app["data_alerts_path"] = "alerts"
+    app["skip_original_data"] = args.get("skip_original_data")
```
hsong-rh, line 117: Suggest to use `args.get("skip_original_data", False)`, feels it's more clear if default is `false`

hsong-rh (dismissed):

plugins/event_source/alertmanager.py
```diff
@@ -2,18 +2,32 @@
...
 Arguments:
-    host: The hostname to listen to. Set to 0.0.0.0 to listen on all
+    host: The webserver hostname to listen to. Set to 0.0.0.0 to listen on all
           interfaces. Defaults to 127.0.0.1
     port: The TCP port to listen to.  Defaults to 5000
+    data_alerts_path: The json path to find alert data. Default to "alerts"
+                      Use empty string "" to treat the whole payload data as
+                      one alert.
+    data_host_path: The json path inside the alert data to find alerting host.
+                    Default to  "labels.instance".
```
Alex-Izquierdo, line 15:
```suggestion
    data_host_path: The json path inside the alert data to find alerting host.
                    Use an empty string `""` to disable the host filtering
                    Default to  "labels.instance".
```

plugins/event_source/alertmanager.py
```diff
@@ -74,14 +109,18 @@ def clean_host(host):
 async def main(queue: asyncio.Queue, args: Dict[str, Any]):
     app = web.Application()
     app["queue"] = queue
+    app["data_host_path"] = args.get("data_host_path", "labels.instance")
+    app["data_path_separator"] = args.get("data_path_separator", ".")
+    app["data_alerts_path"] = args.get("data_alerts_path", "alerts")
```
Alex-Izquierdo, line 114: Type is not enforced. Should try to convert them to string before, otherwise you can have uncontrolled exceptions. 
Only `KeyError` exception is managed.

plugins/event_source/alertmanager.py
bzwei: no need to set empty string if no host can be found. The code automatically skip it.

plugins/event_source/alertmanager.py
bzwei: Do you mean `skip_original_data` in next line? I add `bool()` to convert from string.

Alex-Izquierdo: No, I mean `data_host_path`, `data_path_separator` and `data_alerts_path` If those values are not strings, you will have uncontrolled exceptions from `dpath`.  You must convert them to strings before or check if are strings before calling `dpath.util.get`

plugins/event_source/alertmanager.py
Alex-Izquierdo: The default value is `labels.instance` but the user could want to avoid matching any host at all. This can be done setting the host filtering to empty string `""` and that is why I suggest specifying it in the documentation.

Alex-Izquierdo (changes requested):

Alex-Izquierdo (approved):
lgtm

bzwei:
The main enhancement was already merged through #48. What's left in this PR is some minor fixes and removing an unnecessary loop. A unit test was added too.

Alex-Izquierdo (approved):
lgtm

PR #43, opened 2022-10-31
Add GCP Pub/Sub event source

shopifyalan:


plugins/event_source/gcp_pubsub.py
```diff
@@ -0,0 +1,68 @@
...
+        max_messages: "{{ max_messages }}"
+        retry_deadline: "{{ retry_deadline }}"
+
+"""
+
+import asyncio
+from typing import Any, Dict
+
+from google.api_core import retry
+from google.cloud import pubsub_v1
```
bzwei, line 28: Does pubsub_v1 support asyncio? Typically pull-and-sleep is not the best approach.

Alex-Izquierdo: As far as I see this plugin would block the python thread.[ It could implement futures + callbacks](https://cloud.google.com/pubsub/docs/samples/pubsub-subscriber-async-pull#pubsub_subscriber_async_pull-python) or, I think easier, [use an asyncio library compatible](https://github.com/talkiq/gcloud-aio/tree/master/pubsub)

PR #48, opened 2022-11-03
Dynatrace demo

bzwei:
We can use the existing (enhanced by #42) alertmanager and json_filter to integrate EDA with Dynatrace. Here is the an example on how to monitor the availability of a process and restart it if it crashes.

After some discussion with Ben we felt we should place our demos in this repository. I created demos folder which is excluded from the ansible-galaxy collection installation


Alex-Izquierdo (approved):
LGTM

hsong-rh (approved):

PR #49, opened 2022-11-04
Fix intermittently failing webhook unit test

ttuffin:
Ensure the test always uses ipv4 loopback address instead of 'localhost' to avoid edge cases where the web server and post function will use different versions of IP.
Add empty __init__.py to allow for running test locally.


Alex-Izquierdo (approved):

hsong-rh (approved):
LGTM

PR #52, opened 2022-11-10
Add Github workflow tests for ansible-rulebook installer

ttuffin:
Adding a Github workflow to run ansible-test against the `install_ansible_rulebook` role whenever changes are made. This workflow will cover tests for Debian, RHEL-like and MacOS systems.

Also adding a new test role under `roles/test` which is required by ansible-test.

This PR depends on changes made in #51.

### DIFF
--- a/.github/workflows/installer-tests.yml
+++ b/.github/workflows/installer-tests.yml
@@ -1,0 +1,83 @@
+name: Tests for ansible-rulebook installer
+
+on:
+  pull_request:
+    branches: ["main"]
+    paths:
+      - 'roles/install_ansible_rulebook/**'
+      - 'playbooks/install_rulebook_cli.yml'
+
+jobs:
+  cli-installer-tests:
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - macos-11
+        include:
+          - os: ubuntu-latest
+            container: quay.io/ansible/fedora35-test-container:3.2.0
+            python-version: "3.10"
+          - os: ubuntu-latest
+            container: quay.io/ansible/fedora36-test-container:4.8.0
+            python-version: "3.10"
+          - os: ubuntu-latest
+            container: quay.io/ansible/ubuntu2204-test-container:4.8.0
+            python-version: "3.10"
+          - os: ubuntu-latest
+            container: docker.io/archlinux:latest
+            pkg_install_cmd: pacman -Sy && pacman -S --noconfirm python python-pip python-setuptools
+            ansible_playbook: true
+          - os: ubuntu-latest
+            container: docker.io/debian:latest
+            pkg_install_cmd: apt update && apt install -y init python3 python3-pip python3-setuptools
+            ansible_playbook: true
+          - os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/installer-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+      - 'playbooks/install_rulebook_cli.yml'
+
+jobs:
+  cli-installer-tests:
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+        include:
+          - os: ubuntu-latest
```
Alex-Izquierdo, line 17: wrong value

roles/test/install_ansible_rulebook/test.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Ensure ansible-rulebook can be installed
+  hosts: localhost
+  remote_user: root
+  roles:
+    - install_ansible_rulebook
```
Alex-Izquierdo, line 6: It should run the program to make sure it was installed correctly, isn't?

Alex-Izquierdo (changes requested):

.github/workflows/installer-tests.yml
ttuffin: The `include`? I got that from the [matrix config docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations). It does work, but maybe theres a better way to do it?

Alex-Izquierdo: is not ubuntu, but fedora

Alex-Izquierdo (changes requested):

.github/workflows/installer-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Run ansible-test integration test
+        run: |
+          if [[ "${{ runner.os }}" == "Linux" ]]; then
+            ansible-test integration --docker ${{ matrix.container }} --python 3.10 -v
+          elif [[ "${{ runner.os }}" == "macOS" ]]; then
+            ansible-test integration --local -v
+          fi
+        working-directory: ansible_collections/ansible/eda
+        env:
+          ANSIBLE_TEST_PREFER_PODMAN: 1
```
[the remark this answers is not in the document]
ttuffin: It fails on docker due to this: https://github.com/ansible/ansible/issues/77903

The workers do have podman and it's worked in my test actions.

.github/workflows/installer-tests.yml
Alex-Izquierdo: nevermind, my mistake

Alex-Izquierdo (dismissed):
LGTM

ttuffin:
Seems ansible-test may not fulfill all our requirements, so need to rethink a new implementation.

ttuffin:
Some OS' are covered by `ansible-test` which has pre-built container images. For the remaining OS' that ansible-test does not cover I have added an extra step to manually run `ansible-playbook` against them.

Alex-Izquierdo (approved):
LGTM

Issue #39, opened 2022-10-21
Playbook install-rulebook-cli.yml is poorly named breaking legal ansible-playbook calling conventions


tonykay:
Whilst not a big issue, the use of dashes like in role names should be avoided (IMHO). The use of dashes in the playbook name breaks the ability to follow simple patterns like

```
ansible-galaxy collection install ansible.eda
ansible-playbook ansible.eda.install-rulebook-cli
```

The above fails. However, simply renaming to `install_rulebook_cli.yml` and it succeeds

benthomasson:
Nice catch

ttuffin:
Thanks for the report. We are in the process of refactoring the installer and will fix it there.

PR #53, opened 2022-11-11
Remove dashes from playbooks/install-rulebook-cli.yml

ericzolf:
With the dashes, the direct call with FQCN doesn't work (under Fedora), underscores make it work


ericzolf:
It actually addresses #39

Alex-Izquierdo:
Closed in favor of https://github.com/ansible/event-driven-ansible/pull/51

PR #56, opened 2022-11-15
Override RHEL8 package_list var for CLI installer

ttuffin:
The python3* packages in RHEL8 differ to other OS' in that they point to an older version of Python (3.6). We need to override the package_list var for RHEL8 to ensure that Python3.9 is installed, not 3.6.


Alex-Izquierdo (approved):

ddonahue007 (approved):
lgtm

PR #63, opened 2022-12-19
Remove PIP_NO_BINARY env var

ttuffin:
Closes [AAP-7759](https://issues.redhat.com/browse/AAP-7759). Removing env var from installer playbook since we no longer need to force compilation of JPY from source.

### DIFF
--- a/roles/install_ansible_rulebook/tasks/main.yml
+++ b/roles/install_ansible_rulebook/tasks/main.yml
@@ -49,7 +49,6 @@
 
 - name: Install ansible-rulebook
   environment:
-    PIP_NO_BINARY: jpy
     JAVA_HOME: "{{ java_home_path }}"
   ansible.builtin.pip:
     name: ansible-rulebook


ttuffin:
Tests are failing due to Galaxy rate limiting.

Alex-Izquierdo:
Hi @ttuffin 

> Tests are failing due to Galaxy rate limiting.

We must to install the collection from the git repo instead from galaxy as [we did in ansible-rulebook](https://github.com/ansible/ansible-rulebook/blob/main/.github/workflows/tests.yml#L43). As the playbook should use galaxy by the default, we can move the collection name into a variable, and override it in the test inventory.
https://github.com/ansible/event-driven-ansible/blob/main/roles/install_ansible_rulebook/tasks/main.yml#L59
Something like: 
```
ansible.builtin.command: "ansible-galaxy collection install community.general {{ eda_collection_pkg_name }}"
# test inventory:
# eda_collection_pkg_name: "git+https://github.com/ansible/event-driven-ansible"
```
Not sure if ansible-galaxy will support installing collections from different sources, so you may have to split it in two commands, one for community collection and other one for ansible.eda.

Alex-Izquierdo (approved):

benthomasson (approved):

PR #64, opened 2022-12-19
Allow installation of collection from source

ttuffin:
We want to allow for the installation of the EDA collection from the source repo. This will allow us to work around issues such as hitting Galaxy API limits in our workflows.

### DIFF
--- a/roles/install_ansible_rulebook/tasks/main.yml
+++ b/roles/install_ansible_rulebook/tasks/main.yml
@@ -56,7 +56,7 @@
     state: latest # noqa package-latest
 
 - name: Install collections
-  ansible.builtin.command: ansible-galaxy collection install ansible.eda community.general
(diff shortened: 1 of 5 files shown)


roles/install_ansible_rulebook/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
     state: latest # noqa package-latest
 
 - name: Install collections
-  ansible.builtin.command: ansible-galaxy collection install ansible.eda community.general
+  ansible.builtin.command: ansible-galaxy collection install {{ eda_collection_pkg_name }} community.general
```
Alex-Izquierdo, line 59: Enclose everything in quotes to avoid potential issues with jinja templating.

Alex-Izquierdo (changes requested):

Alex-Izquierdo (approved):

bzwei (approved):

PR #65, opened 2023-01-02
Add cmd source plugin

Alex-Izquierdo:
Cmd source plugin allows the users to run a custom command in background and implement other programming languages as event producers. It can be used in two different ways:

1. In combination with the webhook source plugin to send the events through http.
2. Sending the events through the stdout. Each event must be a JSON serialized string ended with a newline character. With less capabilities at the time to write conditions, if the output is not a JSON string, disabling the `serialize` argument the event will be sent as raw string. 

Examples:
```---
- name: Example with output proccessing
  hosts: all
  sources:
    - ansible.eda.cmd:
        command: node my_custom_plugin.js
        send_output: true
  rules:
    - name: r1
      condition: event.cmd.status == "Running"
      action:
        print_event:
---
- name: Example combined with webhook plugin
  hosts: all
  sources:
    - ansible.eda.webhook:
        port: 6080
    - ansible.eda.cmd:
        command: node my_custom_plugin.js --url http://localhost:6080/eda
  rules:
    - name: r1
      condition: event.payload.status == "Running"
      action:
        print_event:
```


plugins/event_source/cmd.py
```diff
@@ -0,0 +1,92 @@
...
+    if "send_output" in args:
+        send_output = args["send_output"]
+
+    proc = await asyncio.subprocess.create_subprocess_shell(
+        command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+    )
+
+    LOGGER.info("Process started")
+
+    while proc.returncode is None:
```
cutwater, line 79: You can use `proc.wait()` to wait for a process to finish, instead of running a loop with sleep.

cutwater, line 79: You use `PIPE` for both `stdout` and `stderr`  streams, but reading only `stdout` while process is running. This may lead to a deadlock.

Alex-Izquierdo: The loop is necessary because I need to read the stdout while the process is running. The goal is receive each event ([one line](https://github.com/ansible/event-driven-ansible/pull/65/files#diff-0785115d178716a59957832e490c7fc065856df1abe0cc4caa1a117869a64fa2R57)) at a time during the lifetime of the process.

Alex-Izquierdo: Good point. Changed to `communicate()` that according to the documentation prevents the deadlock.

cutwater: It doesn't fix the deadlock issue, because the deadlock appears earlier.

Here is the correct example of handling process `stderr` \ `stdout` stream:

```
    if send_output:
        stdout_mode = asyncio.subprocess.PIPE
    else:
        stdout_mode = asyncio.subprocess.DEVNULL

    proc = await asyncio.subprocess.create_subprocess_shell(
        command, stdout=stdout_mode, stderr=asyncio.subprocess.PIPE
    )
    
    LOGGER.info("Process started")

    if send_output:
        _, stderr = await asyncio.gather(
            process_output(proc, queue),
            proc.stderr.read()
        )
        await proc.wait()
    else:
        _, stderr = await proc.communicate()

    if proc.returncode != 0:
        LOGGER.error(
            "Command failed with code %s: %s", proc.returncode, stderr.decode()
        )

```

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,92 @@
+"""
+cmd.py
+
+An ansible-rulebook event source plugin for running a custom command.
+It can send events to the rule engine from the stdout.
+
+Arguments:
+    command: str
+        The string with the command to execute
```
cutwater, line 9: Is it intended to be an analogue of `ansible.builtin.command` module ?

If yes, the ansible's counterpart expects a string or a list, but regardless of the input format it is processed as a list of arguments, not a shell script.

If no, the documentation is misleading, because it's not a command, but a shell script with the respective rules of escaping arguments, environment variables substitution, etc.

cutwater (changes requested):

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,92 @@
...
+import logging
+from typing import Any, Dict
+
+LOGGER = logging.getLogger("ansible.eda.cmd")
+
+
+async def process_output(
+    proc: asyncio.subprocess.Process, queue: asyncio.Queue
+):
+    # proc.stdout can be None, this is not our case
```
cutwater, line 56: We need a loop here.

```
while line := await proc.stdout.readline():
    ...
```

plugins/event_source/cmd.py
Alex-Izquierdo: I see your point. You are right, I have to deal with stderr always. Your proposal seems more reliable. I will change it and test it later.  Anyway I think I want to bring the stderr in all cases. 

Thank you very much for your suggestions.

plugins/event_source/cmd.py
Alex-Izquierdo: Agree. Changed to process_exec to be a real command.

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,102 @@
...
+    if "send_output" in args:
+        send_output = args["send_output"]
+
+    if send_output:
+        stdout_mode = asyncio.subprocess.PIPE
+
+    cmd_tokens = shlex.split(command)
+    proc = await asyncio.subprocess.create_subprocess_exec(
+        cmd_tokens[0],
+        *cmd_tokens[1:],
```
cutwater, line 80: Minor: Splitting the list of arguments is unnecessary, you simply can unpack it as:

```suggestion
        *cmd_tokens
```

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,102 @@
...
+        stdout=stdout_mode,
+        stderr=asyncio.subprocess.PIPE,
+    )
+
+    LOGGER.info("Process started")
+
+    if send_output:
+        _unused, stderr = await asyncio.gather(
```
cutwater, line 88: Minor: It looks that project's pylint rules can be adjusted. It's completely acceptable to use a single underscore (`_` ) for an unused variables.

plugins/event_source/cmd.py
Alex-Izquierdo: totally true. updated

plugins/event_source/cmd.py
Alex-Izquierdo: Hi @cutwater This is not adjustable. It not comes from pylint directly but by ansible-test which is implemented in the pipelines following the [ansible guidelines](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_testing.html)

It could be ignored adding the proper comment in the code and I thought like you, but I opened a ticket in the community and their reasons have convinced to me: 
https://github.com/ansible/ansible/issues/79646

plugins/event_source/cmd.py
Alex-Izquierdo: changed, please review again

plugins/event_source/cmd.py
cutwater: Ok, makes sense, but according to the explanation provided, it's better to give it a meaningful name. Like `_stdout`.

Alex-Izquierdo: agreed. Updated

cutwater (approved):

benthomasson (commented):
We probably need to add some meta data to track where the events come from.

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,107 @@
...
+    proc: asyncio.subprocess.Process, queue: asyncio.Queue
+):
+    # type hint warn: proc.stdout can be None, this is not our case
+    while output := await proc.stdout.readline():
+        try:
+            event = json.loads(output.decode())
+            await queue.put({"cmd": event})
+        except json.JSONDecodeError as e:
+            LOGGER.error("Can not deserialize JSON data: %s", e)
+        await asyncio.sleep(0)
```
bzwei, line 64: why `sleep(0)` here? line 58 already await.

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,107 @@
...
+async def process_output(
+    proc: asyncio.subprocess.Process, queue: asyncio.Queue
+):
+    # type hint warn: proc.stdout can be None, this is not our case
+    while output := await proc.stdout.readline():
+        try:
+            event = json.loads(output.decode())
+            await queue.put({"cmd": event})
+        except json.JSONDecodeError as e:
+            LOGGER.error("Can not deserialize JSON data: %s", e)
```
bzwei, line 63: Consider simply `put({"cmd": output.decode()})`. The user can then use a customized filter to analyze the string and convert it to a proper json body.

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+async def process_output(
+    proc: asyncio.subprocess.Process, queue: asyncio.Queue
+):
+    # type hint warn: proc.stdout can be None, this is not our case
+    while output := await proc.stdout.readline():
+        try:
+            event = json.loads(output.decode())
+            await queue.put({"cmd": event})
```
bzwei, line 61: You can add a metadata like `{"metadata": {"command": str(args["command"])}}`

bzwei:
@Alex-Izquierdo I don't understand the first use case by sending the cmd output to a web hook. It is a local web hook and what it does is the same as setting `send_output: true`. It does not gain anything for us.

The second recommendation is not to require the cmd to generate a json serializable output. The user can write a customized filter to reformat a string to proper json style. This way the plugin can run any command.

plugins/event_source/cmd.py
```diff
@@ -0,0 +1,107 @@
...
+        await asyncio.sleep(0)
+
+
+async def main(queue: asyncio.Queue, args: Dict[str, Any]):
+    command = str(args["command"])
+    send_output = False
+    stdout_mode = asyncio.subprocess.DEVNULL
+
+    if "send_output" in args:
+        send_output = args["send_output"]
```
bzwei, line 73: should it be `bool(args["send_output"])`?

Alex-Izquierdo: Agree, we can enforce the type. thanks

plugins/event_source/cmd.py
Alex-Izquierdo: Mistake, removed

plugins/event_source/cmd.py
Alex-Izquierdo: I wanted to leave that constraint in the cmd, but I think give the option is not a bad idea. I will add a new option for it

plugins/event_source/cmd.py
Alex-Izquierdo: ok

Alex-Izquierdo:
> @Alex-Izquierdo I don't understand the first use case by sending the cmd output to a web hook. It is a local web hook and what it does is the same as setting `send_output: true`. It does not gain anything for us.
> 
> The second recommendation is not to require the cmd to generate a json serializable output. The user can write a customized filter to reformat a string to proper json style. This way the plugin can run any command.

Hi @bzwei I appreciate your feedback. The first usecase is to reuse some existing webhook (you could have more producers using it). The plugin developer could prefer use always http protocol to send events.  

I think the requirement of a serialized string is not too different than the requirement of a dict for the queue. Deal with structured data it will be always more convenient for users, but I've implemented the optional raw string.

bzwei (dismissed):

Alex-Izquierdo:
> > We probably need to add some meta data to track where the events come from.
> 
> Hi @benthomasson what kind of metadata? Can you provide an example? I thought that should be responsibility of the executed program.

Added the cmdline as metadata @benthomasson

bzwei (dismissed):

ttuffin (approved):

cutwater (approved):

benthomasson:
We should implement specific event source plugins instead of a generic command plugin.

Alex-Izquierdo:
I cannot agree with the reason. The plugin is as generic as webhook is.

PR #67, opened 2023-01-10
Update the URL check test case

ttuffin:
Changing the test case for the URL check source plugin to remove the dependency on the factory and instead place the runner logic in the test case itself. The test has also been extended to check the content of stdout and to include an error handling case when the HTTP server is unreachable.

The same logic will be used in other source plugin test cases so that we can remove the CLI runner factory.


bzwei (approved):

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

ddonahue007 (approved):
lgtm

PR #73, opened 2023-02-02
Remove sleep in webhook integration tests

Alex-Izquierdo:
Fixes flakyness of webhook tests replacing the sleep with a wait for func.

### DIFF
--- a/tests/integration/event_source_webhook/test_webhook_source.py
+++ b/tests/integration/event_source_webhook/test_webhook_source.py
@@ -8,6 +8,20 @@ import subprocess
 from ..utils import CLIRunner, TESTS_PATH
 
 
+def wait_for_events(proc: subprocess.Popen, timeout: float = 5.0):
+    """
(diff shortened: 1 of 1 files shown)


ddonahue007 (dismissed):
lgtm

tests/integration/event_source_webhook/test_webhook_source.py
```diff
@@ -8,6 +8,20 @@
 from ..utils import CLIRunner, TESTS_PATH
 
 
+def wait_for_events(proc: subprocess.Popen, timeout: float = 5.0):
```
bzwei, line 11: why can't you use the same method from utils?

Alex-Izquierdo: It's a mistake, it's should not be there.

bzwei (approved):

ttuffin (approved):

ddonahue007 (approved):
lgtm

PR #71, opened 2023-02-02
Add SQS event source

shanemcd:

PR #76, opened 2023-02-09
Adopt black, flake8, and isort + enforce in CI checks

shanemcd:
Per our discussion after https://github.com/ansible/event-driven-ansible/pull/71#discussion_r1100508257


Alex-Izquierdo:
I personally prefer a limit of 120, but I believe it should be aligned with the 79 chars agreed in the rest of EDA places. I would add the same bunch of linters to the pre-commit config.

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 120
```
benthomasson, line 2: Change length to 88 to match black.

benthomasson (changes requested):

pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[tool.black]
+line-length = 120
```
benthomasson, line 2: 88 is good for normal screens.

benthomasson (changes requested):

benthomasson (approved):

benthomasson (approved):

ttuffin (approved):

PR #77, opened 2023-02-15
Updating plugin to support json payloads from sqs

aknochow:


Alex-Izquierdo (approved):

plugins/event_source/sqs.py
```diff
@@ -50,7 +51,13 @@ async def main(queue: asyncio.Queue, args: Dict[str, Any]):
 
             if "Messages" in response:
                 for msg in response["Messages"]:
-                    await queue.put(msg)
+                    msg_body = msg["Body"]
```
bzwei, line 54: What does a msg typically look like? 
Why msg itself is already a dict but msg["Body"] is a string? 
What are some other keys in msg that can be useful? Should we keep them by putting them to meta field for example?
Why do we demand msg["Body"] to be a json string? Can it be something like "Today is Monday" which is plain text but contains useful info?

plugins/event_source/sqs.py
```diff
@@ -50,7 +51,13 @@ async def main(queue: asyncio.Queue, args: Dict[str, Any]):
...
+                    try:
+                        event = json.loads(msg_body)
+                        await queue.put(event)
+                        await asyncio.sleep(0)
```
bzwei, line 58: no need to sleep

aknochow: will remove that line...

plugins/event_source/sqs.py
aknochow: This is how SQS send its payloads:

image

ReceiptHandle, MD5, and MessageId could be stored as separate meta fields?

This is how the payload comes into AWX now with the proposed changes:
image

bzwei: I think MessageId might be useful. It can be stored in event's meta field.
Is the body example generated by github? So the version, id, etc, are github generated but not sqs generated. Another source may generate a body that is not a json string.

I am thinking this is a generic plugin for all sqs messages we should be able to accept all types. We can try to convert a string to json if possible. Otherwise leave the plain text in the body field.

The event body added to our queue could look like
```
{
  "meta": {"MessageID": "123"},
  "body": "plain text not in json"
}
or
{
  "meta": {"MessageID": "345"},
  "body": {
    "number": 1,
    "pull_request": {}
  }
}
```
`body` may seem redundant in the second example, but you can always set `var_root` in your rulebook to skip it and get the same payload coming to AWX as you showed above.

Alex-Izquierdo: @bzwei  Indeed, the body comes as a string. This PR's is a fix proposed by me to make it working. I don't know in details how it works the library or the sqs service and if this problem can be addressed in another way. @shanemcd probably knows better. This PR's makes usable the plugin as it is now, you can consider it as a bugfix. 

IMO it's acceptable to unblock the work of @aknochow and the plugin can be improved in later iterations.

bzwei: @aknochow @Alex-Izquierdo The change can be as easy as:
```
                for msg in response["Messages"]:
                    meta = {"MessageId": msg["MessageId"]}
                    try:
                        msg_body = json.loads(msg["Body"])
                    except json.JSONDecodeError:
                        msg_body = msg["Body"]

                    await queue.put({"body": msg_body, "meta": meta})
                    await asyncio.sleep(0)
```

plugins/event_source/sqs.py
bzwei: Thanks. sleep(0) makes sense.

plugins/event_source/sqs.py
aknochow: This works... pushing changes...
image

Alex-Izquierdo (approved):

ttuffin (approved):

bzwei (approved):

PR #79, opened 2023-02-17
[AAP-9555] Rename sqs.py to aws_sqs_queue.py

bzwei:
Refactor the arguments

Add unit test

The plugin name and connection argument are aligned with other aws plugins.

### DIFF
--- a/plugins/event_source/aws_sqs_queue.py
+++ b/plugins/event_source/aws_sqs_queue.py
@@ -1,12 +1,22 @@
 """
-sqs.py
+aws_sqs_queue.py
 
 An ansible-rulebook event source plugin for receiving events via an AWS SQS queue.
 
 Arguments:
-    region: AWS region containing the SQS queue.
-    queue:  Name of the SQS queue.
-    wait_seconds: The SQS long polling duration. Set to 0 to disable. Defaults to 2.
+    access_key:    Optional AWS access key ID
+    secret_key:    Optional AWS secret access key
+    session_token: Optional STS session token for use with temporary credentials
+    endpoint_url:  Optional URL to connect to instead of the default AWS endpoints
+    region:        Optional AWS region to use
+    name:          Name of the queue
+    delay_seconds: The SQS long polling duration. Set to 0 to disable. Defaults to 2.
+
+Example:
+    - ansible.eda.aws_sqs:
+        region: us-east-1
+        name: eda
+        delay_seconds: 10
 """
 
 import asyncio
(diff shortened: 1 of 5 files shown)


bzwei:
@aknochow please review

aknochow:
What are your thoughts on using the full queue url just as it would be used in aws api or cli?

What does changing wait_seconds to delay do?  I see both in the api: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html

Is there a reason to put the parameters inside connection or can they all just be at the same level under aws_sqs?

example could look like:
```
    - ansible.eda.aws_sqs:
        queue_url: https://sqs.us-east-2.amazonaws.com/900854079999/gh-pr-webook-queue
        delay_seconds: 2

**Comment by Alex-Izquierdo** [comment]
Hi @aknochow the final url in the end is the result of the combination of multiple parameters (account, queue_name, etc...) even if the library supports a parameter to define it (endpoint_url) is preferable and more consistent leave the composition of the url to the library, based on the parameters. 

@bzwei I imagine you want to homogenize the interfaces of all AWS plugins, but also we could consider the interfaces of all AWS modules for ansible, which don't use the `connection` param, but passing all the supported params at the same level, example: https://docs.ansible.com/ansible/latest/collections/community/aws/sqs_queue_module.html

**Comment by aknochow** [comment]
> 

I was thinking of the ansible sqs modules too (that's where I got queue_url  and delay_seconds from).  They also match the aws api and cli closer.  Can we try to keep the names the same when possible, so region vs region_name? And without the connection: parameter?  What if we take it a step further and name the plugin the same (aws.sqs_queue) and the name would be the queue or you use 'queue'?

examples:

    - ansible.eda.aws.sqs_queue:
        name: gh-pr-webook-queue
        region: us-east-1
        delay_seconds: 5

or leaving the plugin name as is...

    - ansible.eda.aws_sqs:
        queue: gh-pr-webook-queue
        region: us-east-1
        delay_seconds: 5

Thoughts?

**Comment by Alex-Izquierdo** [comment]
@aknochow The name of the plugin should be `ansible.eda.aws_sqs_queue` not `ansible.eda.aws.sqs_queue` IMO
Regardless of that detail I agree with your suggestions.

**Comment by bzwei** [comment]
@aknochow @Alex-Izquierdo I took a subset of the parameters that are in https://docs.ansible.com/ansible/latest/collections/community/aws/sqs_queue_module.html. These are parameters that can be directly passed to the boto library. If we wanted to support the rest parameters we would need to duplicate code from [amazon.aws](https://github.com/ansible-collections/amazon.aws).

File: plugins/event_source/aws_sqs_queue.py [hunk 1]
```diff
@@ -18,16 +28,35 @@
...
+
+
 async def main(queue: asyncio.Queue, args: Dict[str, Any]):
     logger = logging.getLogger()
 
-    region = args.get("region")
-    queue_name = args.get("queue")
-    wait_seconds = int(args.get("wait_seconds", 2))
+    if "name" not in args:
+        raise RuntimeError("Missing queue name")
```
Alex-Izquierdo, line 54: Here is more appropriate the `ValueError` exception: https://docs.python.org/3/library/exceptions.html#ValueError

plugins/event_source/aws_sqs_queue.py
```diff
@@ -18,16 +28,35 @@
...
+    queue_name = args.get("name")
```
Alex-Izquierdo, line 55: Here It would be good to enforce the string type

ttuffin (approved):

PR #80, opened 2023-02-23
Use aws_cloudtrail event source plugin to showcase Ansible Validated Content

alinabuzachis:
Use aws_cloudtrail event source plugin to showcase Ansible Validated Content


bzwei:
This likely needs to be committed to https://github.com/ansible-collections/community.eda which needs to be restructured a bit before accepting new PRs

PR #81, opened 2023-02-23
[AAP-9555] Refactor parameters for aws_cloudtrail

bzwei:
Add unit test

Resolves AAP-9555: [COLLECTION] Unify the param list for aws source plugins

Follow-up #79

### DIFF
--- a/plugins/event_source/aws_cloudtrail.py
+++ b/plugins/event_source/aws_cloudtrail.py
@@ -4,23 +4,24 @@ aws_cloudtrail.py
 An ansible-rulebook event source module for getting events from an AWS CloudTrail
 
 Arguments:
-    connection        - Parameters used to create AWS session
-                        supporters parameters are: region_name, api_version,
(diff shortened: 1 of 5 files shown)


bzwei:
@abikouo  please review

abikouo (dismissed):

plugins/event_source/aws_cloudtrail.py
```diff
@@ -37,6 +38,8 @@
 
 from aiobotocore.session import get_session
 
+from .aws_sqs_queue import connection_args
```
Alex-Izquierdo, line 41: This creates an interdependence between plugins which I think is a bad idea. What do you think moving it to a new file such as aws_utils.py ?

abikouo: I think it is better to do it now, no reason to wait for another aws plugin

Alex-Izquierdo (approved):

abikouo (approved):

ttuffin (approved):

PR #84, opened 2023-03-03
Add a filter that finds hosts and inserts them to meta

bzwei:
This filter can be used together with any source plugins if the event data contains hosts that should be known by ansible-rulebook to run an ansible action.


Alex-Izquierdo (approved):
Very good idea!

I've noticed that for filters the way that args are provided is different than for plugins. I think both should be equal.

mkanoor (approved):

PR #94, opened 2023-03-09
Use absolute path rather than relative path

alinabuzachis:


tests/unit/event_source/test_aws_cloudtrail.py
```diff
@@ -2,11 +2,12 @@
 from unittest.mock import patch
 
 import pytest
+from ansible_collections.ansible.eda.plugins.event_source.aws_cloudtrail import (
```
bzwei, line 5: It is fine to start the import from `plugins`. My local pytest will not even recognize `ansible_collections.ansible.eda`.

plugins/event_source/aws_cloudtrail.py
```diff
@@ -37,8 +37,7 @@
 from typing import Any, Dict
 
 from aiobotocore.session import get_session
-
-from ..utils.aws_utils import connection_args
+from ansible_collections.ansible.eda.plugins.utils.aws_utils import connection_args
```
bzwei, line 40: Let's import from `plugins.utils.aws_utils`

alinabuzachis: Using the relative path will fail when I try to import the source plugin in a rulebook

```
2023-03-08 19:03:01,552 - ansible_rulebook.engine - INFO - load source
2023-03-08 19:03:02,026 - ansible_rulebook.engine - ERROR - Source error
Traceback (most recent call last):
  File "/home/alinabuzachis/tmp/venv-clean/lib64/python3.11/site-packages/ansible_rulebook/engine.py", line 97, in start_source
    module = runpy.run_path(
             ^^^^^^^^^^^^^^^
  File "<frozen runpy>", line 291, in run_path
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File "/home/alinabuzachis/dev/collections/ansible_collections/ansible/eda/plugins/event_source/aws_cloudtrail.py", line 41, in <module>
    from ..utils.aws_utils import connection_args
ImportError: attempted relative import with no known parent package
2023-03-08 19:03:02,035 - ansible_rulebook.engine - INFO - Start a playbook action runner for Rules for cloud.aws_ops to ensure the cloudtrail is encrypted
```

bzwei (approved):

tests/unit/event_source/test_aws_cloudtrail.py
bzwei: it seems we do need a full path. Thank you.

ttuffin (approved):

PR #101, opened 2023-03-13
Kafka plugin accepts non json string

bzwei:
Parse json data if possible, otherwise leave it as pure string. Always place the kafa message under key body in the event placed in the queue.

Resolves AAP-9944: Kafka source plugin only supports json


Alex-Izquierdo (approved):
lgtm

ttuffin (approved):

Alex-Izquierdo (approved):

PR #106, opened 2023-03-15
Add security support to webhook plugin

bzwei:
Client needs to send an bearer authorization token. Enable TLS support if a certificate file is provided.

Resolves AAP-4826: Add security check for webhook and alert manger's local server

We may make TLS default in a future PR.

### DIFF
--- a/extensions/eda/plugins/event_source/webhook.py
+++ b/extensions/eda/plugins/event_source/webhook.py
@@ -4,17 +4,24 @@ webhook.py
 An ansible-rulebook event source module for receiving events via a webhook.
 
 Arguments:
-    host: The hostname to listen to. Set to 0.0.0.0 to listen on all
-          interfaces. Defaults to 127.0.0.1
-    port: The TCP port to listen to.  Defaults to 5000
+    host:     The hostname to listen to. Set to 0.0.0.0 to listen on all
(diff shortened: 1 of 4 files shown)


ttuffin (dismissed):
Tested, works ok. In a follow up PR it might be good to catch SSL exceptions to provide a clearer message, i.e when the cert and key do not match.

plugins/event_source/webhook.py
```diff
@@ -4,49 +4,86 @@
...
+        raise web.HTTPUnauthorized(text="Invalid authorization token")
+
+    return await handler(request)
+
+
 async def main(queue: asyncio.Queue, args: Dict[str, Any]):
-    app = web.Application()
+    app = web.Application(middlewares=[bearer_auth])
     app["queue"] = queue
+    app["token"] = args["token"]
```
Alex-Izquierdo, line 61: I really believe this param should be optional. First to keep the backward compatibility and also not all the clients are able to send the token.

Alex-Izquierdo (changes requested):

Alex-Izquierdo (dismissed):
lgtm

hsong-rh (approved):

ttuffin (approved):

PR #109, opened 2023-03-22
[AAP-10473] Add a filter to insert source info

mkanoor:
The event payload can be embellished with source name and type under the meta key. It allows for tracking of the origin of the event

https://issues.redhat.com/browse/AAP-10473


mkanoor:
Made this into a builtin filter

PR #117, opened 2023-04-24
Adding nats eda source plugin

burnyd:


Alex-Izquierdo:
Hi @burnyd 

We want to express our sincere appreciation for your effort and contribution. After careful consideration, we have concluded that this plugin doesn't align with the scope of our official collection, and unfortunately, it will not be included.

However, this does not diminish the value of your work. We believe your plugin can still be highly beneficial and encourage you to offer it through your own collection. You are welcome to upload it [to galaxy](https://galaxy.ansible.com/) with the "eda" tag.

Thank you for your understanding and contribution.

PR #121, opened 2023-04-26
Added an event filter to normalize keys in the event payload

mkanoor:
The keys in the event payload should follow the convention of of an ansible variable name. A key name can only consist of alpha numeric and under score charachters. All non conforming charachters are converted to _.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#creating-valid-variable-names

This is a generic form of the dashes_to_underscore filter, which only translates dashes. This filter translates all non conforming characters

### DIFF
--- a/extensions/eda/plugins/event_filter/normalize_keys.py
+++ b/extensions/eda/plugins/event_filter/normalize_keys.py
@@ -1,0 +1,69 @@
+"""
+normalize_keys.py:
+    An event filter that changes keys that contain non alpha numeric or
+    underscore to undersocres.
+    For instance, the key server-name becomes the new key server_name
+    If there are consecutive non alpa numeric or under score, they would
+    be coalesced into a single underscore
+    For instance the key server.com/&abc becomes server_com_abc
+    instead of server_com__abc
+
+    If there is a existing key with the normalized name, it will get overwritten
+    by default. If you don't want to over write it you can pass in overwrite: False
+    The default value of overwrite is True.
+
+Arguments:
+    * overwrite: Overwrite the values if there is a collision with a new key.
+
+Usage in a rulebook, a filter is usually attached to a source in the rulebook:
+    example1:
+    ....
+    sources:
+      - my_source:
+         ....
+        filters:
+          ansible.eda.normalize_keys:
+
+
+    example2: To prevent over writing of keys
+    ....
+    sources:
+      - my_source:
+         ....
+        filters:
+          ansible.eda.normalize_keys:
+            overwrite: False
+
+"""
+
+import multiprocessing as mp
+import re
+
+normalize_regex = re.compile("[^0-9a-zA-Z_]+")
+
+
+def main(event, overwrite=True):
+    logger = mp.get_logger()
+    logger.info("normalize_keys")
+    return _normalize_embedded_keys(event, overwrite, logger)
+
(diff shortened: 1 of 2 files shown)


extensions/eda/plugins/event_filter/normalize_keys.py
```diff
@@ -0,0 +1,50 @@
...
+normalize_keys.py:
+    An event filter that changes keys that contain non alpha numeric or
+    underscore to undersocres.
+    For instance, the key X-Y becomes the new key X_Y.
+    If there are consecutive non alpa numeric or under score, they would
+    be coalesced into a single underscore
+    For instance the key X/?Y becomes X_Y instead of X__Y
+
+    If there is a existing key with the normalized name, you can pass
+    in the overwrite argument to overwrite it with the new key
```
Alex-Izquierdo, line 11: We should say that overwrite=false would remove the key in case of collision.

extensions/eda/plugins/event_filter/normalize_keys.py
```diff
@@ -0,0 +1,50 @@
...
+
+def main(event, overwrite=True):
+    logger = mp.get_logger()
+    logger.info("normalize_keys")
+    q = []
+    q.append(event)
+    regex = re.compile("[^0-9a-zA-Z_]+")
+    while q:
+        o = q.pop()
+        if isinstance(o, dict):
```
Alex-Izquierdo, line 36: We are skipping lists of dicts, don't?

extensions/eda/plugins/event_filter/normalize_keys.py
```diff
@@ -0,0 +1,50 @@
...
+                new_key = regex.sub("_", key)
+                if new_key != key:
+                    del o[key]
+                    if new_key in o and overwrite:
+                        o[new_key] = value
+                        logger.debug("Replacing %s with %s", key, new_key)
+                    elif new_key not in o:
+                        o[new_key] = value
+                        logger.debug("Replacing %s with %s", key, new_key)
+
```
Alex-Izquierdo, line 49: Probably have a log message indicating that the key is permanently removed if overwrite=false can be a good idea.

Alex-Izquierdo (changes requested):
It would be possible to use a more descriptive variable names instead of letters? Readability is value.

extensions/eda/plugins/event_filter/normalize_keys.py
mkanoor: @Alex-Izquierdo A single event is always a dict its never an array.

mkanoor:
@Alex-Izquierdo Please re review I have made the suggested changes

extensions/eda/plugins/event_filter/normalize_keys.py
Alex-Izquierdo: A single event can contain a list of dicts. 
```{ "name": "alex", "phones": [{"home": 777777}, {"work": 5555555}]}```

Alex-Izquierdo (changes requested):

mkanoor:
@Alex-Izquierdo Please re-review

bzwei (approved):

Alex-Izquierdo (approved):

PR #123, opened 2023-05-04
Fix Webhook source without token

vkrizan:
Fixes Webhook event source in case the token isn't used/set.
Otherwise it raises `KeyError`.


vkrizan:
cc @bzwei

ttuffin (approved):

hendersonreed (approved):

Issue #37, opened 2022-10-20
Add a Kubernetes Events source


joshgav:
In addition to resource-related events as described in #36, Kubernetes provides a stream of Events which contain info about cluster and workload state and changes. Some details here: <https://www.cncf.io/blog/2021/12/21/extracting-value-from-the-kubernetes-events-feed/>

Many of those events are great candidates for alerts and remediation via Ansible.

Could you create a Kubernetes "Events" source too? Thanks!

Issue #36, opened 2022-10-20
Add a Kubernetes resource source


joshgav:
Kubernetes provides a stream of resource create/update/delete events via its API as described [here](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

If Ansible could subscribe to this stream of events it could act as an out-of-band controller/operator. Similar concepts can be found in [metacontroller](https://metacontroller.github.io/metacontroller/) and [keptn](https://keptn.sh/docs/concepts/architecture/).

To this end, could you please create a "source" based on this stream of Kubernetes resource-related events? Perhaps it could be configurable re which resource types it watches. Thank you!

---

NOTE: the stream of resource-related events is _not_ the same as Kubernetes Events - capital E - which are a specific type of resource. Those would be nice to subscribe to as well but as another source; here's an article with more about them: <https://www.cncf.io/blog/2021/12/21/extracting-value-from-the-kubernetes-events-feed/>

yajo:
If this were implemented, basically it would deprecate [the whole Operator SDK for Ansible](https://sdk.operatorframework.io/docs/building-operators/ansible/quickstart/), which is way more alien for Ansible devs than EDA.

joshgav:
This work from @sabre1041 seems to address this issue: https://www.ansible.com/blog/kubernetes-meets-event-driven-ansible

Thanks Andy! :partying_face:

joshgav:
Hi @sabre1041, one thing I think I noted in reading your article is that we don't get all the properties of the Kubernetes resource itself in the rulebook. That is, we don't get the actual properties of the `ConfigMap`, `Deployment` or whatever CR was created. Those properties describe the desired state of the resource so they will often be needed by the reacting playbook.

The pattern in Go operators is almost always to start the reconciliation loop by retrieving the current resource from the API server. That is, after the operator is notified that a resource named `x` was ADDED/MODIFIED/DELETED it then goes and retrieves what exactly `x` currently is per the API server.

Perhaps in the future if users desire it we could enable this event source to make that callback automatically and "hydrate" all the resource properties in advance before handing it off to the playbook.

ccamacho:
Hi @joshgav and @yajo I have been playing with EDA+Kubernetes and this might be interesting https://github.com/ansible/event-driven-ansible/pull/133

[...]

PR #134, opened 2023-06-13
Include a new Kubernetes event source

ccamacho:
This commit includes a new events source
plugin for monitoring changes in a Kubernetes cluster. This plugin enables the usage of any API class
from the Kubernetes Python client.

Closes: #36
Closes: #37

### DIFF
--- a/demos/dynatrace-demo/instructions.md
+++ b/demos/dynatrace-demo/instructions.md
@@ -1,7 +1,7 @@
 # Ansible Rulebook + Dynatrace DEMO
 
 ## Description
-In this demo we will register a host to Dynatrace and set up a webhook to send
+In this demo, we will register a host to Dynatrace and set up a webhook to send
 problem notifications to ansible-rulebook CLI. Dynatrace monitors the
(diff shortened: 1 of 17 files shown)


the-it-jaeger:
I think there is a definite need for this. For example, I would like to have Ansible react when OCP ClusterAutoscaler adds or removes new Machines/Nodes by adding (or removing) their IP addresses to (or from) firewalls and load balancers.

Alex-Izquierdo:
Hi @ccamacho 

We want to express our sincere appreciation for your effort and contribution. After careful consideration, we have concluded that this plugin doesn't align with the scope of our official collection, and unfortunately, it will not be included. It might be a better fit for a dedicated collection of curated plugins for Kubernetes.

However, this does not diminish the value of your work. We believe your plugin can still be highly beneficial and encourage you to offer it through your own collection. You are welcome to upload it [to galaxy](https://galaxy.ansible.com/) with the "eda" tag.

As a fellow Red Hatter, we also suggest reaching out to your product management team to explore if such a collection could be supported within Red Hat as a maintained collection.

Thank you for your understanding and contribution.

PR #135, opened 2023-06-15
Add tox.yml and update tox.ini based on eda-partner-testing

eclarizio:
We needed to update the tox files as a requirement for certification

### DIFF
--- a/extensions/eda/plugins/event_source/file.py
+++ b/extensions/eda/plugins/event_source/file.py
@@ -22,7 +22,8 @@ from watchdog.events import RegexMatchingEventHandler
 from watchdog.observers import Observer
 
 
-def send_facts(queue, filename):
+def send_facts(queue, filename: str) -> None:
+    """Send facts to the queue."""
     with open(filename) as f:
         data = yaml.safe_load(f.read())
         if data is None:
@@ -44,7 +45,8 @@ def send_facts(queue, filename):
                 queue.put(item)
 
 
-def main(queue, args):
+def main(queue, args: dict) -> None:
+    """Load facts from YAML files initially and when the file changes."""
     files = [os.path.abspath(f) for f in args.get("files", [])]
 
     if not files:
@@ -57,18 +59,18 @@ def main(queue, args):
         def __init__(self, **kwargs) -> None:
             RegexMatchingEventHandler.__init__(self, **kwargs)
 
-        def on_created(self, event):
+        def on_created(self, event: dict) -> None:
             if event.src_path in files:
                 send_facts(queue, event.src_path)
 
-        def on_deleted(self, event):
+        def on_deleted(self, event: dict) -> None:
             pass
 
-        def on_modified(self, event):
+        def on_modified(self, event: dict) -> None:
             if event.src_path in files:
                 send_facts(queue, event.src_path)
 
-        def on_moved(self, event):
+        def on_moved(self, event: dict) -> None:
             pass
 
     observer = Observer()
@@ -87,9 +89,13 @@ def main(queue, args):
 
 
 if __name__ == "__main__":
+    """MockQueue if running directly."""
 
     class MockQueue:
-        def put(self, event):
-            print(event)
+        """A fake queue."""
+
+        async def put(self, event: dict) -> None:
+            """Print the event."""
+            print(event) # noqa: T201
 
     main(MockQueue(), {"files": ["facts.yml"]})
--- a/extensions/eda/plugins/event_source/file.py
+++ b/extensions/eda/plugins/event_source/file.py
@@ -15,16 +15,16 @@ Example:
 
 """
 
-import os
+import pathlib
 
 import yaml
 from watchdog.events import RegexMatchingEventHandler
 from watchdog.observers import Observer
 
 
-def send_facts(queue, filename: str) -> None:
+def send_facts(queue, filename: str) -> None:  # noqa: ANN001
     """Send facts to the queue."""
-    with open(filename) as f:
+    with pathlib.Path(filename).open() as f:
         data = yaml.safe_load(f.read())
         if data is None:
             return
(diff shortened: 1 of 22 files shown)


Alex-Izquierdo (dismissed):
I wonder if there is an specific action for tox, anyway LTGM

Alex-Izquierdo:
mmm, seems that ruff is raising issues.

eclarizio:
@Alex-Izquierdo Yeah, I think we will have to fix all of the issues in this PR since with the new files it's throwing a ton of errors

ccamacho:
This change looks like a nice thing to have, maybe due that this will be a fairly huge change, it might be good to split it in multiple PRs to avoid regressions and making it easy to review.

Instead of calling all the environments at once in tox, you could make it pass step by step running the environments separately.

eclarizio:
> This change looks like a nice thing to have, maybe due that this will be a fairly huge change, it might be good to split it in multiple PRs to avoid regressions and making it easy to review.
> 
> Instead of calling all the environments at once in tox, you could make it pass step by step running the environments separately.

I don't disagree, though there's going to be a ton of changes either way, unfortunately :(. Still, it's probably a good idea to change this one to only be the `ruff` issues, and make subsequent PRs for `darglint`, `pylint-event-source`, and `pylint-event-filter`.

Alex-Izquierdo:
Splitting the tox envs in multiple actions it would the best as well to split the updates to the code for an easier review.

ccamacho:
> Splitting the tox envs in multiple actions it would the best as well to split the updates to the code for an easier review.

I have something for that from a while a go.. but didnt finish it, let me see if I can find it. Also the tox.ini should not be in the .github/workflows folder.

eclarizio:
> > Splitting the tox envs in multiple actions it would the best as well to split the updates to the code for an easier review.
> 
> I have something for that from a while a go.. but didnt finish it, let me see if I can find it. Also the tox.ini should not be in the .github/workflows folder.

I'm simply following the instructions from https://github.com/ansible/eda-partner-testing/blob/main/README.md#toxini which says it should be in that folder.

eclarizio:
Alright, I've updated this PR to just run the ruff check, which should make this a bit more manageable. Future PRs will enable the others that we need (`darglint`, `pylint-event-source`, and `pylint-event-filter`)

I'm going to need some help to resolve all of these as for some of them I just don't have the knowledge. If you're tagged here, please take a look at the changes made to the file that I did as some of them were done automatically via a `--fix` helper command and others were done manually so they may be incorrect or odd. Note all of these changes were done just to appease the tox ruff gods, so if it's breaking other stuff, I will need help to fix that as well while keeping ruff happy.

I've also created a commit for each file as most of them had remaining issues, and I listed the remaining issues I need assistance with in the commit message, hopefully that helps make it a bit easier to isolate, though I acknowledge it's probably a lot of noise and can be condensed later.

Here's a list of all the files that need to be adjusted, along with the majority committer(s).
`extensions/eda/plugins/event_filter/dashes_to_underscores.py`: @benthomasson 
`extensions/eda/plugins/event_filter/normalize_keys.py`: @mkanoor 
`extensions/eda/plugins/event_source/alertmanager.py`: @bzwei @hsong-rh 
`extensions/eda/plugins/event_source/aws_cloudtrail.py`: @bzwei @abikouo 
`extensions/eda/plugins/event_source/aws_sqs_queue.py`: @bzwei @shanemcd
`extensions/eda/plugins/event_source/azure_service_bus.py`: @bzwei @benthomasson 
`extensions/eda/plugins/event_source/file.py`: @benthomasson 
`extensions/eda/plugins/event_source/file_watch.py`: @benthomasson 
`extensions/eda/plugins/event_source/generic.py`: @ttuffin 
`extensions/eda/plugins/event_source/journald.py`: @konstruktoid 
`extensions/eda/plugins/event_source/kafka.py`: @bzwei 
`extensions/eda/plugins/event_source/range.py`: @benthomasson 
`extensions/eda/plugins/event_source/tick.py`: @benthomasson 
`extensions/eda/plugins/event_source/url_check.py`: @benthomasson 
`extensions/eda/plugins/event_source/webhook.py`: @bzwei

ccamacho:
something like this @eclarizio https://github.com/ansible/event-driven-ansible/pull/138 which will enable all the people who wants to make the lint checks consistent without breaking any job.

ccamacho:
> ly following the instructions from https://github.com/ansible/eda-partner-testing/blob/main/README.md#toxini which says it should be in tha

I think that is a reference guide for the tests that should be executed, for example the tox.ini file is not exclusive of github actions so it should not be there. If a partner uses i.e. GitLab, Zuul, Jenkins, or Travis this might not apply.

bzwei:
@eclarizio I will address most of the errors

extensions/eda/plugins/event_source/file.py
```diff
@@ -15,69 +15,71 @@
...
 
     for filename in files:
         send_facts(queue, filename)
+    _observe_files(queue, files)
+
 
+def _observe_files(queue, files: list[str]) -> None: # noqa: ANN001
     class Handler(RegexMatchingEventHandler):
-        def __init__(self, **kwargs) -> None:
+        def __init__(self: "Handler", **kwargs) -> None: # noqa: ANN003
```
eclarizio, line 61: Can we not use the same type as here https://github.com/ansible/event-driven-ansible/pull/135/commits/e7f4d98cc526840acd086542fb99a28d6a1ad178#diff-a614b4efc2dd1f32c82962f861ab13d16ac3a843d8b2c38a6136b21dcc3a8421R37?

bzwei: I believe this file is not compatible with ansible-rulebook engine. It should not be used. I put some simple fix to suppress the warning. This file should be removed or rewritten.

bzwei:
@eclarizio The test failed with python3.8 because syntax like dict[str, Any] is not supported. Should we fix them or just drop version 3.8?

eclarizio:
@bzwei @Alex-Izquierdo I think I would propose we drop 3.8. [According to this link](https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle#rhel8_application_streams) Python 3.9 in RHEL is going EOL in May 2024, and Python 3.8 is already not supported. That's for RHEL 8, and for RHEL 9 it looks like we want to be using 3.11.

Alex-Izquierdo:
At some point I remember to remove or propose the removal of python3.8 but @mkanoor had some reason to keep it. I don't remember well. For me, totally agree to remove it.

bzwei (approved):

hsong-rh (approved):

Issue #145, opened 2023-06-30
False positives in tox checks


ccamacho:
The way we run the tox linters makes them to display a false positive output, all the errors are not displayed.

ssbarnea:
The report is quite vague but I do fully agree that the way tox is configured to run is... far from optimal. I will work these days to address it.

PR #146, opened 2023-06-30
False positive output in tox linter checks

ccamacho:
This commit:
- Makes easier to test individually each commit by allowing to run the GitHub actions on each fork push and pull request.
- Remove the appended path in the ruff tox check that breaks it to parse recursivelly all the python code.

TODO: This should be applied to all the other tox check to make them show correctly all the linters errors.

Closes: #145

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,9 +1,7 @@
 name: Run integration tests
 on:
   push:
-    branches: ["main"]
   pull_request:
-    branches: ["main"]
   # Run the integration tests every 8 hours.
   # This will help to identify faster if
   # there is a CI failure related to a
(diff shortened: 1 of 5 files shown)


ccamacho:
This is expected to make the tox check to fail, additional PRs should be submitted fo fix all the lint nits.

ttuffin:
I am not sure about running all tests for every push to all branches, it could get a little noisy in my opinion. @Alex-Izquierdo, thoughts?

ccamacho:
> I am not sure about running all tests for every push to all branches, it could get a little noisy in my opinion. @Alex-Izquierdo, thoughts?

@ttuffin thanks for the feedback!! Yeah, the thing is that in your fork, the tests will run whenever you create a branch, make your changes, and testing them before pushing a PR. Also we are not running any check i.e. when merging things in stable/1.0 which might be useful to detect things that might break there.

Alex-Izquierdo:
Hi @ccamacho thanks for the contribution, unfortunately there are still checks in ruff that are not passing.

Issue #153, opened 2023-07-24
systemd-python not available on pip


thedoubl3j:
the systemd-python package is not available via pypi. the package being listed in the requirements.txt causes EE builds to break due to it not being packaged correctly there. It is available via dnf per the packages [documentation](https://github.com/systemd/python-systemd#installation). 

With this package being available in dnf, it will need to be moved to a bindep.txt file for proper consumption by ansible-builder so that it can resolve the dependency correctly. 

```
+ /usr/bin/python3.9 -m pip install --cache-dir=/output/wheels -r /tmp/src/requirements.txt
WARNING: Running pip install with root privileges is generally not a good idea. Try `python3.9 -m pip install --user` instead.
Collecting azure_servicebus
  Downloading azure_servicebus-7.11.1-py3-none-any.whl (390 kB)
Collecting aiobotocore
  Downloading aiobotocore-2.5.2-py3-none-any.whl (72 kB)
Collecting aiohttp
  Downloading aiohttp-3.8.5-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.1 MB)
Collecting aiokafka
  Downloading aiokafka-0.8.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.1 MB)
Collecting watchdog
  Downloading watchdog-3.0.0-py3-none-manylinux2014_x86_64.whl (82 kB)
Collecting systemd_python
  Downloading systemd-python-235.tar.gz (61 kB)
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3.9 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-iv9vdn8z/systemd-python/setup.py'"'"'; __file__='"'"'/tmp/pip-install-iv9vdn8z/systemd-python/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-zvaexoht
         cwd: /tmp/pip-install-iv9vdn8z/systemd-python/
    Complete output (15 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-iv9vdn8z/systemd-python/setup.py", line 55, in <module>
[...]

PR #154, opened 2023-07-24
remove systemd-python from pypi requirements and list in bindep

thedoubl3j:
systemd-python is not available on pip, this causes issues when a user attempts to build an EE from the collection. the package is available via dnf so moving the dependency to a bindep.txt will pass the dep to dnf during the ansible-builder build phase which will be able to resolve the dep. see #153 for more context on the issue


benthomasson (approved):

bzwei (approved):

Alex-Izquierdo (changes requested):
I don't think to say that systemd-package is unavailable on pypi or wrong packaged is correct. As is described in the docs, it just  needs to have installed several system dependencies. 

As this process is complicated, I agree with remove it from the requirements.txt. But the readme must indicate that `python3-systemd` package must be installed in addition to the requirements.txt.

bindep.txt
```diff
@@ -0,0 +1 @@
+systemd-python
```
Alex-Izquierdo, line 1: According [to the documentation](https://github.com/systemd/python-systemd#installation) and my own tests, the package name is `python3-systemd`

Alex-Izquierdo (changes requested):

Alex-Izquierdo:
Hi @thedoubl3j thanks for your contribution and update the right package name. Readme needs to be modified too, to indicate that `python3-systemd` package must be installed, otherwise the users that follow these instructions would have unexpected errors.

thedoubl3j:
@Alex-Izquierdo done and squashed commits as well

README.md
```diff
@@ -23,6 +23,12 @@ The python module dependencies are not installed by ansible-galaxy. They must be
 pip install -r requirements.txt
 ```
 
+The binary module dependencies are not installed by ansible-galaxy. They must also  be installed manually using dnf:
```
konstruktoid, line 26: Write a generic instruction or add additional instructions using other package managers as well?

README.md
```diff
@@ -23,6 +23,12 @@ The python module dependencies are not installed by ansible-galaxy. They must be
...
+
+```
+dnf install python3-systemd
+```
+
```
Alex-Izquierdo, line 31: IMO this test is clearer:
```suggestion
To use the journald source plugin, it is also necessary to have the system package named `python3-systemd` installed on Debian or Fedora operating systems. For other platforms, loot at the [systemd-python documentation](https://github.com/systemd/python-systemd#installation). 
```

Alex-Izquierdo: Hi @thedoubl3j Do you agree to commit my suggestion?

Alex-Izquierdo:
package python3-systemd must be installed as well in the CI workflows to pass the tests.

ssbarnea:
While use of `bindep.txt` seems nice (i wanted to add it myself), I am closing this because referred bug was incorrect. Note that I am in favour of adding a bindep.txt fine, but if we do, it must be accompanied by some GHA/tox changes that would be testing it for supported platforms.

Issue #156, opened 2023-07-31
webhook source plugins returns an uncontrolled 500 exception when the request doesn't contain a valid JSON body


Alex-Izquierdo:
## How to reproduce:
- Run a basic rulebook with the webhook source plugin. 
- Run a POST request with a body that is not a valid JSON string. 

## Actual result:

```
curl http://localhost:5001/ -d 'not a json'
```

response:
```
500 Internal Server Error
```

Rulebook output:
```
2023-07-31 13:38:59,962 - aiohttp.server - ERROR - Error handling request
Traceback (most recent call last):
  File "/home/alex/repos/redhat/ansible/eda/ansible-rulebook/venv/lib64/python3.9/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
  File "/home/alex/repos/redhat/ansible/eda/ansible-rulebook/venv/lib64/python3.9/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/home/alex/.ansible/collections/ansible_collections/ansible/eda/extensions/eda/plugins/event_source/webhook.py", line 30, in webhook
    payload = await request.json()
  File "/home/alex/repos/redhat/ansible/eda/ansible-rulebook/venv/lib64/python3.9/site-packages/aiohttp/web_request.py", line 671, in json
    return loads(body)
  File "/usr/lib64/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

```

## Expected result:

Webhook plugin should handle this situation either returning an 400 with an exception indicating that only json is supported or to return an 200 accepting a non-json body.

Alex-Izquierdo:
Internal ref: https://issues.redhat.com/browse/AAP-14569

PR #157, opened 2023-08-02
Fix invalid json response for webhook plugin

Alex-Izquierdo:
Fixes https://github.com/ansible/event-driven-ansible/issues/156
Jira: https://issues.redhat.com/browse/AAP-14569

### DIFF
--- a/extensions/eda/plugins/event_source/webhook.py
+++ b/extensions/eda/plugins/event_source/webhook.py
@@ -34,8 +34,8 @@ async def webhook(request: web.Request) -> web.Response:
     try:
         payload = await request.json()
     except json.JSONDecodeError as e:
-        logger.error("Wrong body request: failed to decode JSON payload: %s", e)
-        raise web.HTTPBadRequest(text="Invalid JSON payload")
+        logger.warning("Wrong body request: failed to decode JSON payload: %s", e)
+        raise web.HTTPBadRequest(text="Invalid JSON payload") from None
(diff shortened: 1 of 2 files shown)


Dostonbek1 (approved):

extensions/eda/plugins/event_source/webhook.py
```diff
@@ -29,7 +31,11 @@
 @routes.post(r"/{endpoint:.*}")
 async def webhook(request: web.Request) -> web.Response:
     """Return response to webhook request."""
-    payload = await request.json()
+    try:
+        payload = await request.json()
+    except json.JSONDecodeError as e:
+        logger.warning("Wrong body request: failed to decode JSON payload: %s", e)
+        raise web.HTTPBadRequest(text="Invalid JSON payload") from None
```
hsong-rh, line 38: Why do we need raise exception from `None`?

Alex-Izquierdo: Hi @hsong-rh This is to pass ruff checks. When you raise an different exception in a try block, ruff enforce to use the "raise from" to have the whole trace. In this case we don't want it because there is no any info to provide for debugging, so it must be "None"

Ref: https://beta.ruff.rs/docs/rules/raise-without-from-inside-except/

hsong-rh (approved):

bzwei (approved):

PR #169, opened 2023-08-22
Fix aws_cloudtrail even duplication bug

hakbailey:
Fixes a minor bug in the aws_cloudtrail event_source plugin where cloudtrail events were retrieved and added to queue twice. The boto3 client `StartTime` parameter needs to be set to the last retrieved event time before the next batch of events is retrieved instead of after.


Alex-Izquierdo (approved):
LGTM

bzwei (approved):

PR #171, opened 2023-09-14
Added kafka SCRAM authentication based extension (Work in progress)

ritzshah:
I see that there is kafka extension which works with SSL but it doesn't work with SCRAM authentication. So I created this one and tested to be working hence thought of raising a PR to see if this can be added to the collections.

Here's an example I used to test as a rulebook in EDA
https://github.com/ritzshah/eda-rulebooks/blob/main/rulebooks/hello.yml

Here's the image to use when creating a decision environment to test.
https://quay.io/repository/rshah/ansible-rulebook-kafka-scram


Alex-Izquierdo:
Thanks for your contribution @ritzshah, but it looks like we would do the same thing that already does the current plugin but with a different authentication. I'm not sure if it is justified to have two different plugins for the same thing instead of update the current one to support different authentication ways. Also it should have tests to cover this use case. We need to update our documentation and pipelines to ensure new additions are properly covered by tests.

ritzshah:
Cleaning up with a new PR.

PR #173, opened 2023-09-22
Extension for SCRAM based authentication for Kafka connectivity. This includes SSL as well.

ritzshah:
Added Kafka extension with SCRAM authentication. You can also use SSL instead of SCRAM with this extension. 

This is tested and the image is located here : 
https://quay.io/repository/rshah/ansible-rulebook-kafka-scram-aio

The rule book example is located here : 
https://github.com/ritzshah/eda-rulebooks/blob/main/rulebooks/chat-survey.yml

### DIFF
--- a/extensions/eda/plugins/event_source/kafka_scram_aio.py
+++ b/extensions/eda/plugins/event_source/kafka_scram_aio.py
@@ -1,0 +1,144 @@
+"""kafka_scram_aio.py.
+
+An ansible-rulebook event source plugin for receiving events via a kafka topic.
+
+Arguments:
+---------
+    host:      The host where the kafka topic is hosted
+    port:      The port where the kafka server is listening
+    cafile:    The optional certificate authority file path containing certificates
+               used to sign kafka broker certificates
+    certfile:  The optional client certificate file path containing the client
+               certificate, as well as CA certificates needed to establish
+               the certificate's authenticity
+    keyfile:   The optional client key file path containing the client private key
+    password:  The optional password to be used when loading the certificate chain
+    check_hostname:  Enable SSL hostname verification. [True (default), False]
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
Thank you very much for your contribution @ritzshah, but I think it is not justified to have a new plugin for kafka instead of update the existing one. Also it should have tests for the new options.

ritzshah:
@Alex-Izquierdo 

Thanks for reviewing and your comments. I have tested this particular one in an openshift environment. I see that the existing plugin kafka.py is connecting to kafka, 
1) without user/password, 
2) SSL

Adding a 3rd option for SCRAM was becoming a little difficult in the same plugin and the plugin options were becoming more complex if someone wants to add all options in a rulebook, so I thought maybe like the file plugin kafka can have multiple plugin. 

Also can you clarify what do you mean b test options.

extensions/eda/plugins/event_source/kafka_scram_aio.py
```diff
@@ -0,0 +1,144 @@
...
+        bootstrap_servers=f"{host}:{port}",
+        group_id=group_id,
+        enable_auto_commit=True,
+        max_poll_records=1,
+        auto_offset_reset=offset,
+        security_protocol="SSL" if cafile else security_protocol,
+        ssl_context=context if cafile else None,
+        sasl_plain_password=sasl_plain_password,
+        sasl_plain_username=sasl_plain_username,
+        sasl_mechanism = None if cafile else sasl_mechanism
```
bzwei, line 105: What is the sasl_mechanism if secuirty_protocol is SASL_SSL?

bzwei:
Closing this one since #177 is merged.

PR #175, opened 2023-10-03
add eda modules to the collection

jainnikhil30:
This PR adds 5 modules to the eda collection:

- activation
- credential
- decision_environment
- project
- user

### DIFF
--- a/plugins/modules/activation.py
+++ b/plugins/modules/activation.py
@@ -10,15 +10,14 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 ANSIBLE_METADATA = {
-    "metadata_version": "1.1",
     "status": ["preview"],
     "supported_by": "community",
 }
 
-DOCUMENTATION = """
+DOCUMENTATION = '''
 ---
 module: activation
-author: Nikhil Jain
+author: "Nikhil Jain"
 short_description: Create, restart or delete rulebook activations in EDA Controller.
 description:
   - This module allows you to create, restart or delete activations in a EDA controller.
@@ -68,12 +67,12 @@ options:
     description:
       - Desired state of the resource.
     default: "present"
-    choices: ["present", "absent", "exists", "restart"]
+    choices: ["present", "absent", "exists"]
     type: str
 extends_documentation_fragment: ansible.eda.auth
-"""
+'''
 
-EXAMPLES = """
+EXAMPLES = '''
 - name: Create EDA Activation
   ansible.eda.activation:
     name: "Example Activation"
@@ -92,7 +91,7 @@ EXAMPLES = """
     decision_environment: "Default Decision Environment"
     enabled: False
     state: absent
-"""
+'''
 
 
 from ..module_utils.eda_controller_api import EDAControllerAPIModule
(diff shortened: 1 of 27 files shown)


plugins/modules/activation.py
```diff
@@ -0,0 +1,202 @@
...
+    type: str
+    required: true
+  extra_vars:
+    description:
+      - The extra variables for the rulebook activation, default is ''.
+    type: str
+    default: ''
+  restart_policy:
+    description:
+      - The restart policy for the rulebook activation, default is 'always'.
```
bzwei, line 52: default is on-failure

plugins/modules/activation.py
```diff
@@ -0,0 +1,202 @@
...
+def main():
+    argument_spec = dict(
+        name=dict(type="str", required=True),
+        description=dict(required=False),
+        project=dict(type="str", required=True),
+        rulebook=dict(type="str", required=True),
+        extra_vars=dict(type="str", default=""),
+        restart_policy=dict(
+            type="str",
+            default="always",
```
bzwei, line 109: default="on-failure"

plugins/modules/activation.py
```diff
@@ -0,0 +1,202 @@
...
+        activation_fields["rulebook_id"] = rulebook_id
+
+    if decision_environment_id is not None:
+        # this is resolved earlier, so save an API call and don't do it again in the loop above
+        activation_fields["decision_environment_id"] = decision_environment_id
+
+    if restart_policy is not None:
+        activation_fields["restart_policy"] = restart_policy
+
+    if restart_policy is not None:
```
bzwei, line 185: duplicated code from line 182

plugins/modules/activation.py
```diff
@@ -0,0 +1,202 @@
...
+ANSIBLE_METADATA = {
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = """
+---
+module: activation
+author: "Nikhil Jain (@jainnikhil30)"
+short_description: Create, restart or delete rulebook activations in EDA Controller.
```
bzwei, line 21: Other than create and delete, no other operation such as restart, stop, disable, enable an existing activation is supported in this PR.

plugins/modules/activation.py
```diff
@@ -0,0 +1,202 @@
...
+
+- name: Delete EDA Activation
+  ansible.eda.activation:
+    name: "Example Activation"
+    description: "Example Activation description"
+    project: "Example Project"
+    rulebook: "hello_controller.yml"
+    decision_environment: "Default Decision Environment"
+    enabled: False
+    state: absent
```
bzwei, line 93: For absent only name field is required, why do you need to add all other fields in the example? The same for other modules such as credential, decision_environment, project, and user.

plugins/modules/activation.py
jainnikhil30: Sure, made the change

plugins/modules/activation.py
jainnikhil30: Sure, made the change

plugins/modules/activation.py
jainnikhil30: My bad, removed it.

plugins/modules/activation.py
jainnikhil30: That is correct changed the description

plugins/modules/activation.py
jainnikhil30: Yeah the reason I had done it that way, I kept those other fields as required fields. Now I have removed it and you can delete all objects with just name and state as "absent"

bzwei:
LGTM. @jainnikhil30 Please fix the length violations. Is it possible to add tests for the modules?

bzwei:
@jainnikhil30 please resubmit the PR to branch api-module. We're not ready to merge it to main yet. Thanks.

jainnikhil30:
@bzwei Okay, does that mean the modules wont be part of the actual collection release ? If yes, then how would users use the modules ?

jainnikhil30:
Closed this in favor of #178

PR #176, opened 2023-10-12
Add template for event emitter plugins

benthomasson:
This is an example of an event emitter plugin.


kaiokmo:
@benthomasson I think we can get this one merged for the sake of having a simple example of how a source plugin is written.

WDYT? Should we undraft it?

PR #177, opened 2023-10-23
feat: Add Kafka SASL_SSL authentication

jouir:
Add `security_protocol`, `sasl_mechanism`, `sasl_plain_username` and `sasl_plain_password` arguments like aiokafka.AIOKafkaConsumer client.

https://aiokafka.readthedocs.io/en/stable/consumer.html

### DIFF
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -14,11 +14,21 @@ Arguments:
     keyfile:   The optional client key file path containing the client private key
     password:  The optional password to be used when loading the certificate chain
     check_hostname:  Enable SSL hostname verification. [True (default), False]
+    verify_mode: Whether to try to verify other peers' certificates and how to
+               behave if verification fails. [CERT_NONE, CERT_OPTIONAL,
+               CERT_REQUIRED (default)]
     encoding:  Message encoding scheme. Default to utf-8
     topic:     The kafka topic
     group_id:  A kafka group id
     offset:    Where to automatically reset the offset. [latest, earliest]
                Default to latest
+    security_protocol: Protocol used to communicate with brokers. [PLAINTEXT, SSL,
+               SASL_PLAINTEXT, SASL_SSL]. Default to PLAINTEXT
+    sasl_mechanism: Authentication mechanism when security_protocol is configured.
+               [PLAIN, GSSAPI, SCRAM-SHA-256, SCRAM-SHA-512, OAUTHBEARER].
+               Default to PLAIN.
+    sasl_plain_username: Username for SASL PLAIN authentication
+    sasl_plain_password: Password for SASL PLAIN authentication
 
 
 
@@ -27,6 +37,7 @@ Arguments:
 import asyncio
 import json
 import logging
+from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
 from typing import Any
 
 from aiokafka import AIOKafkaConsumer
@@ -48,22 +59,38 @@ async def main(  # pylint: disable=R0914
     keyfile = args.get("keyfile")
     password = args.get("password")
     check_hostname = args.get("check_hostname", True)
+    verify_mode = args.get("verify_mode", "CERT_REQUIRED")
     group_id = args.get("group_id", None)
     offset = args.get("offset", "latest")
     encoding = args.get("encoding", "utf-8")
+    security_protocol = args.get("security_protocol", "PLAINTEXT")
 
     if offset not in ("latest", "earliest"):
         msg = f"Invalid offset option: {offset}"
         raise ValueError(msg)
 
-    if cafile:
-        context = create_ssl_context(
+    verify_modes = {
+        "CERT_NONE": CERT_NONE,
+        "CERT_OPTIONAL": CERT_OPTIONAL,
+        "CERT_REQUIRED": CERT_REQUIRED,
+    }
+    try:
+        verify_mode = verify_modes[verify_mode]
+    except KeyError as exc:
+        msg = f"Invalid verify_mode option: {verify_mode}"
+        raise ValueError(msg) from exc
+
+    ssl_context = None
+    if cafile or security_protocol.endswith("SSL"):
+        security_protocol = security_protocol.replace("PLAINTEXT", "SSL")
+        ssl_context = create_ssl_context(
             cafile=cafile,
             certfile=certfile,
             keyfile=keyfile,
             password=password,
         )
-        context.check_hostname = check_hostname
+        ssl_context.check_hostname = check_hostname
+        ssl_context.verify_mode = verify_mode
 
     kafka_consumer = AIOKafkaConsumer(
         topic,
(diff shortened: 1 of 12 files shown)


jouir:
Same as #173 but it updates the existing kafka event source instead. Not tested with SCAM.

Alex-Izquierdo (commented):
Hello @jouir thank you very much for your contribution. I have some concerns about how you are managing the options. 
1. The security_protocol should not be overridden to "SSL" if its value is "SASL_SSL." It requires a different set of options (including those defined in the sasl_mechanism). Additionally, it seems to be missing an option: "SASL_PLAINTEXT."
2. The ssl_context is required for SSL and SASL_SSL options. Therefore, for our arguments, cafile is mandatory in such cases and should be enforced. Otherwise, the error returned by aiokafka could be confusing for users.
3. It would be beneficial to check all the matrices of required options on our side or capture "ValueError" exceptions from the Kafka client to provide nice and meaningful messages to the users, rather than relying on the default traceback output.

The pull request would be merged more quickly if it includes tests that deploy the required Kafka with Docker Compose, as we already do in the integration tests. Otherwise, we will have to test it manually, which will take more time.

jouir:
Hello @Alex-Izquierdo, thank you for the review. I appreciate!

>     1. The security_protocol should not be overridden to "SSL" if its value is "SASL_SSL." It requires a different set of options (including those defined in the sasl_mechanism). 

The `security_protocol` is overridden only if there is a `cafile` and the value is `PLAINTEXT`, like before.

```
>>> args = {"security_protocol": "SASL_SSL"}
>>> security_protocol = args.get("security_protocol", "PLAINTEXT")
>>> cafile = args.get("cafile")
>>> if cafile or security_protocol in ["SSL", "SASL_SSL"]:
...     if security_protocol == "PLAINTEXT":
...             security_protocol = "SSL"
... 
>>> security_protocol
'SASL_SSL'
```

> Additionally, it seems to be missing an option: "SASL_PLAINTEXT."

Good point. I will add it, thanks.

>     2. The ssl_context is required for SSL and SASL_SSL options. Therefore, for our arguments, cafile is mandatory in such cases and should be enforced. Otherwise, the error returned by aiokafka could be confusing for users.

The `cafile` is not required to create a `SSLContext`. We can see in the [aiokafka documentation](https://aiokafka.readthedocs.io/en/stable/api.html#aiokafka.helpers.create_ssl_context) that `cafile` defaults to `None`:

> If CA not specified (by either cafile, capath, cadata) default system CA will be used if found by OpenSSL.

Same with [ssl](https://docs.python.org/3/library/ssl.html#ssl.create_default_context):

> cafile, capath, cadata represent optional CA certificates to trust for certificate verification, as in SSLContext.load_verify_locations(). If all three are None, this function can choose to trust the system’s default CA certificates instead.

Which is the case by default here. We use such setup to use very basic encryption without error from aiokafka nor ssl.

>     3. It would be beneficial to check all the matrices of required options on our side or capture "ValueError" exceptions from the Kafka client to provide nice and meaningful messages to the users, rather than relying on the default traceback output.

As of today, I have not encountered any `ValueError` exception with error messages nor traceback thrown by aiokafka because of a misconfiguration. We can provide examples to guide users, but testing all the possible cases would duplicate the work already performed by aiokafka.

> The pull request would be merged more quickly if it includes tests that deploy the required Kafka with Docker Compose, as we already do in the integration tests. Otherwise, we will have to test it manually, which will take more time.

Sure. I will try to provide a Kafka broker with SASL auth matching my own case using docker-compose and associated tests.

Have a nice day

Alex-Izquierdo:
Hello again @jouir Thanks for your quickly response! :)
You are right regarding the behavior of the SSL option. I may have misunderstood the documentation. Sorry for the confusion. I await for the SASL_PLAINTEXT and some test. Thanks a lot.

jouir:
I have pushed a new commit including tests for SASL_PLAINTEXT and SASL_SSL authentications, with self-signed certificates.

The condition to enable the SSLContext has been changed to check for security_protocol ending with SSL. But if you prefer the old condition based on an explicit list of security protocols (like `SSL`, `SASL_SSL`) instead, I can revert the change.

Integration tests were testing the `ansible.eda.kafka` official source instead of the one of the repository so I've added the sources path to the `CLIRunner` to test current changes. I had this issue running tests locally. I don't know if there is the same issue running on the CI/CD infrastructure.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -51,19 +58,26 @@ async def main(  # pylint: disable=R0914
...
 
     if offset not in ("latest", "earliest"):
         msg = f"Invalid offset option: {offset}"
         raise ValueError(msg)
 
-    if cafile:
-        context = create_ssl_context(
+    ssl_context = None
+    if cafile or security_protocol.endswith("SSL"):
+        if security_protocol == "PLAINTEXT":
```
jouir, line 72: `security_protocol` should be `SASL_SSL` when `cafile` is provided and `security_protocol` is `SASL_PLAINTEXT`.

tests/integration/event_source_kafka/test_kafka_rules.yml
```diff
@@ -1,7 +1,7 @@
 - name: test kafka source plugin
   hosts: localhost
   sources:
-    - ansible.eda.kafka:
+    - kafka:
```
Alex-Izquierdo, line 4: I'm not sure why you have removed the FQCN here, but "sources" option, even if it is tracked in the CLIRunner is only for development purposes and should not be used in these tests, can you come back to "ansible.eda.kafka" ?

jouir: When I run tests locally with the FQCN (`ansible.eda.kafka`) and the default sources directory (`None`), they all fail with the following error:

```
2023-10-27 10:16:20,963 - ansible_rulebook.engine - ERROR - Source error Could not find source plugin for ansible.ed...ble.eda.kafka
2023-10-27 10:16:20,980 - ansible_rulebook.cli - ERROR - Terminating One of the source plugins failed
```

When I add the `-S extensions/eda/plugins/event_source` argument, they fail too with the same error.

Tests succeed only if the Kafka source plugin is renamed to `kafka` (without FQCN) and with the overridden sources directory.

Do you have a solution?

Alex-Izquierdo: Hi @jouir certainly is weird, It should work. Have you check in the collection is properly installed before running ansible-rulebook? The pipeline should not fail. 
https://github.com/ansible/event-driven-ansible/blob/main/.github/workflows/integration-tests.yml#L43

jouir: Adding the collection locally using ansible-galaxy was the trick. Thanks!

jouir:
My new revision fixes the following issues:
* Removed useless variable assignments from the main function because the complexity of the function was too high (> 50 statements)
* Replaced a single quote that was copy/paste from [ssl](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_mode) documentation, because it wasn't `'` or `"`
* Use FQCN to test the collection
* Re-ordered the imports

Alex-Izquierdo (approved):
Thank you very much @jouir 
There is an error raised by flake, other than that it looks great to me.

jouir:
An unused line was left over. It's now removed.

Alex-Izquierdo (approved):

bzwei (approved):

PR #189, opened 2024-01-28
Fix spacing as expected by black

ericzolf:
black 24.1.0 has introduced at least one new spacing rule and the pipeline errors were driving me crazy so here a fix.


ttuffin (approved):

mkanoor (approved):

ttuffin:
Thanks for the contribution @ericzolf

PR #191, opened 2024-01-31
Added integration test for project module

sarathbrp:


kdelee:
guessing that either `api-module` may need rebased on devel/main or maybe you should rebase on `api-module`

PR #197, opened 2024-02-16
feat: add raise_error and log_error option to insert_hosts_to_meta filter

bzwei:
https://issues.redhat.com/browse/AAP-17232

### DIFF
--- a/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
+++ b/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
@@ -14,6 +14,14 @@ Arguments:
                     string but contains multiple hosts, use this parameter to
                     delimits the hosts. Treat the vale as a single host if the
                     parameter is not present.
+    raise_error:    Whether raise PathNotExistError if host_path does not
+                    exist in the event. Default to false.
+                    It is recommended to turn it on during the rulebook
+                    development time. You can then turn it off for production.
+    log_error:      Whether log an error message if host_path does not
+                    exist in the event. Default to true.
+                    You can turn if off if it is expected to have events not
+                    having the host_path to avoid noises in the log.
 
 Example:
 -------
@@ -22,14 +30,23 @@ Example:
           host_path: "app.target"
           path_separator: "."
           host_separator: ";"
+          raise_error: true
+          log_error: true
 
 """
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 import dpath
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PathNotExistError(Exception):
+    """Cannot find the path in the event."""
 
 
 def main(
@@ -37,6 +54,9 @@ def main(
     host_path: str | None = None,
     host_separator: str | None = None,
     path_separator: str = ".",
+    *,
+    raise_error: bool = False,
+    log_error: bool = True,
 ) -> dict[str, Any]:
     """Extract hosts from event data and insert into meta dict."""
     if not host_path:
@@ -44,8 +64,13 @@ def main(
 
     try:
         hosts = dpath.get(event, host_path, path_separator)
-    except KeyError:
+    except KeyError as error:
         # does not contain host
+        msg = f"Event {event} does not contain {host_path}"
+        if log_error:
+            LOGGER.error(msg)  # noqa: TRY400
+        if raise_error:
+            raise PathNotExistError(msg) from error
         return event
 
     if isinstance(hosts, str):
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo (approved):

Alex-Izquierdo (dismissed):

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -32,20 +35,29 @@
...
     """Extract hosts from event data and insert into meta dict."""
     if not host_path:
         return event
 
     try:
         hosts = dpath.get(event, host_path, path_separator)
-    except KeyError:
+    except KeyError as error:
         # does not contain host
+        if raise_error:
```
mkanoor, line 58: @bzwei Should we also log a warning message here irrespective of the raise_error setting?

mkanoor:
@bzwei should we add documentation here too https://ansible.readthedocs.io/projects/rulebook/en/latest/filters.html

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -14,6 +14,8 @@
                     string but contains multiple hosts, use this parameter to
                     delimits the hosts. Treat the vale as a single host if the
                     parameter is not present.
+    raise_error:    Whether raise PathNotExistError if host_path does not
+                     exist in the event. Default to false
```
mkanoor, line 18: @bzwei I think we should mention that while developing the rulebook its recommended that you set the raise_error to True and once you are content that it works you can set it to false.

bzwei:
> @bzwei should we add documentation here too https://ansible.readthedocs.io/projects/rulebook/en/latest/filters.html

Yes.

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
bzwei: It may be normal for events not containing a host. This may generate a lot of noises in the log.

Alex-Izquierdo: Agree. Maybe INFO level.

Alex-Izquierdo:
I'm concerned about mixing scopes of the documentation. Why not create a `docs/event_filter/insert_docs_to_meta.md` which was the plan from the beginning and we can continue writing documentation for the rest of things?

Alex-Izquierdo (dismissed):

mkanoor:
@bzwei 
With this PR changes and the issue that we had seen with the broken path
```
 host_path: event.payload.meta.hosts
```

Nothing changes, there is no log messages and there is no exception. A customer has to be told to use the new raise_exceptions option to see anything meaningful. We are back to square one. This PR doesn't help the customer self diagnose the issue. They might go back to using the limit approach.

We need to log.error by default so out of the box we see an error in the Activation log that the customer can be told to turn off by modifying the rulebook which is an easier change for them.

Maybe log_errors: True|False.  True by default.

If they feel there is too much noise they could just set the log_errors: False

Alex-Izquierdo (dismissed):

Alex-Izquierdo:
Hi @bzwei Ruff is failing with:
```
extensions/eda/plugins/event_filter/insert_hosts_to_meta.py:52:5: PLR0913 Too many arguments to function call (6 > 5)
extensions/eda/plugins/event_filter/insert_hosts_to_meta.py:71:13: TRY400 Use `logging.exception` instead of `logging.error`
```
CI criteria for eda collections is a pending discussion.

In the case of PLR0913 you already have a exception but I think it must be in line #52. Anyway, since filters receives the plain args and generally speaking, I think it is perfectly reasonable to add the exception for PLR0913 in general in https://github.com/ansible/event-driven-ansible/blob/main/tox.ini#L18

In the case of TRY400 I think we want to keep the check but this case justifies a line exception.

Alex-Izquierdo (approved):

Alex-Izquierdo (dismissed):

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -14,6 +14,14 @@
                     string but contains multiple hosts, use this parameter to
                     delimits the hosts. Treat the vale as a single host if the
                     parameter is not present.
+    raise_error:    Whether raise PathNotExistError if host_path does not
+                    exist in the event. Default to false.
+                    It is recommended to turn it on during the rulebook
+                    development time. You can then turn it off for production.
+    log_error:      Whether raise PathNotExistError if host_path does not
```
mkanoor, line 21: @bzwei I think this should be 
```
Log an error message if the path is missing, by default we will always log an error. If the user is convinced that the path is valid and only some of incoming events have the host path, they can turn off the error logging by setting log_error to False.
```

mkanoor (approved):

Alex-Izquierdo (approved):

PR #205, opened 2024-05-16
feat: enable webhook mTLS support

bzwei:
The webhook plugin already accepts a certificate to enable TLS. Add new options to accept CA certificates to enable mTLS to validate client certificates.

AAPRFE-1252: Add mTLS authentication for webhook event source plugin in Event-Driven Ansible


extensions/eda/plugins/event_source/webhook.py
```diff
@@ -18,6 +18,10 @@
     certfile: The optional path to a certificate file to enable TLS support
     keyfile:  The optional path to a key file to be used together with certfile
     password: The optional password to be used when loading the certificate chain
+    cafile:   The optional path to a file containing CA certificates used to validate
+              clients' certificates
+    capath:   The optional path to a directory containing CA certificates
+              Provide either cafile or capth to enable mTLS support
```
Alex-Izquierdo, line 24: typo
```suggestion
              Provide either cafile or capath to enable mTLS support
```

extensions/eda/plugins/event_source/webhook.py
```diff
@@ -127,6 +131,30 @@ async def hmac_verify(request: web.Request, handler: Callable) -> web.StreamResp
     return await handler(request)
 
 
+def _get_ssl_context(args: dict[str, Any]) -> ssl.SSLContext:
```
Alex-Izquierdo, line 134: Minor, type hint is wrong, it can also return `None`

eclarizio (dismissed):

extensions/eda/plugins/event_source/webhook.py
```diff
@@ -30,18 +34,23 @@
...
 import json
 import logging
 import ssl
-from collections.abc import Callable
+import typing
 from typing import Any
 
 from aiohttp import web
 
+if typing.TYPE_CHECKING:
```
benthomasson, line 51: Why is this check needed?

mkanoor:
@bzwei Can we add a test for this in this file https://github.com/ansible/event-driven-ansible/blob/9fc76cd4aeef528d7e6c523a56666f04b5fe8ea4/tests/unit/event_source/test_webhook.py#L52
We can create a cert for the server side and a cert for the client side. When we do the mTLS we need to pass in the CAfile to the server side which should be the client cert.

benthomasson (dismissed):

bzwei:
> @bzwei Can we add a test for this in this file
> 
> https://github.com/ansible/event-driven-ansible/blob/9fc76cd4aeef528d7e6c523a56666f04b5fe8ea4/tests/unit/event_source/test_webhook.py#L52
> 
> 
> We can create a cert for the server side and a cert for the client side. When we do the mTLS we need to pass in the CAfile to the server side which should be the client cert.

mTLS unit test is enabled

mkanoor:
@bzwei we need to mark these files in the exception list similar to  https://github.com/ansible/eda-server/blob/main/.gitleaks.toml

bzwei:
> @bzwei we need to mark these files in the exception list similar to https://github.com/ansible/eda-server/blob/main/.gitleaks.toml

added

mkanoor (approved):

benthomasson (approved):

Issue #207, opened 2024-05-20
Include headers in kafka message


clyang82:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/ansible-rulebook/issues) for duplicates.
- [X] I understand that ansible-rulebook is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

We use EDA to connect with kafka and the kafka message is following the [cloudevents](https://cloudevents.io/) specification. the [example](https://github.com/cloudevents/spec/blob/v1.0/spec.md#example) of cloudevents is 
```
{
    "specversion" : "1.0",
    "type" : "com.github.pull.create",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}
```
currently, EDA only handles the `data` field. Do not support `headers`.

### Steps to reproduce

It is very simple to connect kafka as source of EDA.

### Current results

```
{'hosts': ['all'],
 'inventory': '../inventory',
 'project_data_file': None,
 'rule': 'Notify',
 'rule_run_at': '2024-05-20T12:36:10.706124Z',
 'rule_set': 'Read messages from a kafka topic',
 'rule_set_uuid': 'c30c6b82-1331-4bb6-b7a6-963068415dea',
 'rule_uuid': '9612402d-5720-455f-b6d4-cff2fc9b22a1',
 'variables': {'CAFILE_PATH': 'ca.crt',
               'CERTFILE_PATH': 'client.crt',
               'HOST': 'xxx',
               'KEYFILE_PATH': 'client.key',
               'TOPIC': 'event',
               'event': {'body': [{'clusterId': 'dfd94785-6296-48a7-afab-2c72e9ccf816',
                                   'createdAt': '2024-05-20T08:50:58Z',
                                   'eventName': 'demo-managed-a1-03.17cea',
                                   'eventNamespace': 'demo-managed-a1',
                                   'message': 'demo-managed-a1 is available '
                                              'now',
                                   'reason': '',
                                   'reportingController': 'cluster-import-controller',
[...]

PR #208, opened 2024-05-20
Support headers

clyang82:
fixed: #207

### DIFF
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -129,7 +129,8 @@ async def receive_msg(
         # Process headers
         try:
             headers = {header[0]: header[1].decode(encoding) for header in msg.headers}
-            event["headers"] = headers
+            event["meta"] = {}
+            event["meta"]["headers"] = headers
         except UnicodeError:
             logger.exception("Unicode error while decoding headers")
 
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -29,6 +29,7 @@ Arguments:
                Default to PLAIN.
     sasl_plain_username: Username for SASL PLAIN authentication
     sasl_plain_password: Password for SASL PLAIN authentication
+    with_headers: Include message headers. [True, False (default)]
 
 
 
@@ -64,6 +65,7 @@ async def main(  # pylint: disable=R0914
     offset = args.get("offset", "latest")
     encoding = args.get("encoding", "utf-8")
     security_protocol = args.get("security_protocol", "PLAINTEXT")
+    with_headers = args.get("with_headers", False)
 
     if offset not in ("latest", "earliest"):
         msg = f"Invalid offset option: {offset}"
@@ -110,6 +112,7 @@ async def main(  # pylint: disable=R0914
 
     try:
         async for msg in kafka_consumer:
+            event = {}
             data = None
             try:
                 value = msg.value.decode(encoding)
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo (commented):
Hi @clyang82 Thanks for your contribution. 

I think there is no need to make it dependent on an option in the plugin and we could just send the headers always. What do you think @bzwei?

Tests should be updated to check that we send correctly the headers in the event.

clyang82:
I am afraid sending the headers always may break the backward compatibility if someone is dealing with the whole event, including `body` and `meta`.

Alex-Izquierdo:
> I am afraid sending the headers always may break the backward compatibility if someone is dealing with the whole event, including `body` and `meta`.

I think there is no risk at all, because we are extending the existing schema of the event with a new root key of the event, preserving the existing "body" key. To me this is a bug and we should send the headers always. 

cc @bzwei @mkanoor

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -119,8 +122,24 @@ async def main(  # pylint: disable=R0914
...
                 logger.exception("Unicode Error")
 
+            if with_headers:
+                headers = None
+                try:
+                    headers_dict = {
+                        header[0]: header[1].decode(encoding) for header in msg.headers
+                    }
+                    headers_json = json.dumps(headers_dict)
+                    headers = json.loads(headers_json)
```
Alex-Izquierdo, line 132: Why do you serialize and deserialize the data? It looks like an unnecessary use of cycles per event.  
Since you are composing a python dict, we could send it directly into event["headers"]

clyang82:
> > I am afraid sending the headers always may break the backward compatibility if someone is dealing with the whole event, including `body` and `meta`.
> 
> I think there is no risk at all, because we are extending the existing schema of the event with a new root key of the event, preserving the existing "body" key. To me this is a bug and we should send the headers always.
> 
> cc @bzwei @mkanoor

Thanks for your comments. Remove the option to send the headers always.

extensions/eda/plugins/event_source/kafka.py
clyang82: I am new for python. Yes. do not need to serialize and deserialize the data. Updated.

Alex-Izquierdo (approved):

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -109,24 +109,49 @@ async def main(  # pylint: disable=R0914
...
+            event["headers"] = headers
+        except UnicodeError:
+            logger.exception("Unicode error while decoding headers")
+
+        # Process message body
+        try:
+            value = msg.value.decode(encoding)
+            data = json.loads(value)
+        except json.decoder.JSONDecodeError:
+            logger.exception("JSON decode error, storing raw value")
```
Alex-Izquierdo, line 141: This is an expected usecase, we don't want `logger.exception`, much better INFO.

Alex-Izquierdo (approved):

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -109,24 +109,49 @@ async def main(  # pylint: disable=R0914
...
+    """Receive messages from the Kafka topic and put them into the queue."""
+    logger = logging.getLogger()
+
+    async for msg in kafka_consumer:
+        event = {}
+
+        # Process headers
+        try:
+            headers = {header[0]: header[1].decode(encoding) for header in msg.headers}
+            event["headers"] = headers
```
bzwei, line 132: Can we place headers under meta like https://github.com/ansible/event-driven-ansible/blob/main/extensions/eda/plugins/event_source/webhook.py#L71?

clyang82: yes. that makes sense. When discussing metadata broadly, headers can be one of the elements included in the metadata. Thanks.

Alex-Izquierdo (approved):

PR #209, opened 2024-05-21
Add API modules

kdelee:
This adds modules for interacting with the EDA API
Includes tests and github action for running them

It needs a rebase


eclarizio:
@bzwei You originally requested this work to be merged into the `api-module` branch instead of main, is there any reason this can't be brought to the main branch now?

.github/workflows/tests.yml
```diff
@@ -45,3 +45,54 @@ jobs:
...
+
+      - name: Checkout ansible/eda-server
+        uses: actions/checkout@v3
+        with:
+          repository: ansible/eda-server
+          path: eda-server # Note: Specify a separate path for this repository
+       
+      - name: Run API in background
+        run: |
+          docker compose -p eda -f docker-compose-stage.yaml pull
```
Alex-Izquierdo, line 79: This docker-compose used by default the latest build from main. This works for now, but we will need to release tagged images of eda-server and bind that tag to the version of the collection, otherwise the tests would fail in the future with breaking changes.

Alex-Izquierdo:
Obsolete, new work is going to be added.

PR #215, opened 2024-07-13
feat: generic source supports events in yaml file

mkanoor:
This feature allows you to store an array of events in a yaml file and load the events from the yaml file when the rulebook is running. The new argument is called **payload_file**

### DIFF
--- a/extensions/eda/plugins/event_source/generic.py
+++ b/extensions/eda/plugins/event_source/generic.py
@@ -4,6 +4,7 @@ The event data to insert into the queue is specified in the required
 parameter payload and is an array of events.
 
 Optional Parameters:
+payload_file A yaml with an array of events can be used instead of payload
 randomize    True|False Randomize the events in the payload, default False
 display      True|False Display the event data in stdout, default False
 timestamp    True|False Add an event timestamp, default False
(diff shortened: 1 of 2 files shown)


extensions/eda/plugins/event_source/generic.py
```diff
@@ -99,6 +104,10 @@ def __init__(self: Generic, queue: asyncio.Queue, args: dict[str, Any]) -> None:
         """Insert event data into the queue."""
         self.queue = queue
         field_names = [f.name for f in fields(Args)]
+
+        if "payload_file" in args:
+            args["payload"] = ""
```
Alex-Izquierdo, line 109: This should not be an empty array? Which also makes me notice that the type of "payload" is wrong in the dataclass.

Alex-Izquierdo (approved):

extensions/eda/plugins/event_source/generic.py
mkanoor: It can be a simple string which gets converted to an array.

Alex-Izquierdo: ok, I see.

bzwei (approved):

PR #217, opened 2024-08-02
chore: fix spellcheck errors and enable cspell

Akasurde:
##### SUMMARY

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE

- Docs Pull Request

Partial: #219

PR #218, opened 2024-08-02
test: fix pre-commit reported issues

ssbarnea:
Blocks: #217


.pre-commit-config.yaml
```diff
@@ -13,6 +18,7 @@ repos:
         name: Run ansible-test sanity
         entry: ansible-test sanity
         require_serial: true
+        stages: [manual]
```
ssbarnea, line 21: If anyone wonders, these tasks should not be here, too slow and already broken on both pre-commit ci and locally. We will ensure we run them using tox.

Akasurde (approved):

Issue #219, opened 2024-08-02
Refactor testing: tox and github actions usage to simplify testing


ssbarnea:
This is an "epic" level ticket for tracking future changes to how we run tests locally on CI (GHA), so we can ensure we have a much better experience with both.

- [x] enable pre-commit.ci
- [x] configure tox `lint` job to run pre-commit
- [x] add spellchecker to pre-commit (cspell and/or codespell)
- [x] move linters supported by pre-commit inside pre-commit as it is much easier to maintain (and faster)
- [x] enable py311 and py312 GHA pipelines
- [ ] enable one macos aarch64 job (uncovers dependency issues on arm)
- [ ] enable ansible version matrix for testing with tox (via tox-ansible probably)
- [x] refactor GHA execution to make use of reusable tox workflows. Example https://github.com/ansible/ansible-lint/blob/main/.github/workflows/tox.yml#L29-L44 (lower the maintenance needed on GHA)
- [ ] report coverage using codecov.io

## Optional (less urgent)

- [ ] enable depentabot for actions
- [ ] enable use of constraints and depentabot updates for them

AAP-28987

PR #221, opened 2024-08-03
Upgrade linters and refactor how they run

ssbarnea:
- Address two bugs uncovered by newer versions of the linters
- Configure flake8, black, isort, pylint to run as pre-commit hooks
- Avoid running tox from within pre-commit hook by making that hook manual (pre-commit was not designed to be used this way)
- Ensure pre-commit runs on CI via both pre-commit.ci app check and our own GHA job (as they complement each other, only first being able to propose auto-fixes)

Partial: #219


ssbarnea:
I will be merging this to unlock follow-up work. Note that integration tests failures are unrelated to this change.

PR #227, opened 2024-08-06
Replace unmaintained docker-compose with podman-compose

ssbarnea:
As python based docker-compose is unmaintained and with last release more than 3 year old, we make use of podman-compose instead, which is a wrapper that can make use of any docker-compose if found, in addition to native podman-compose tool.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,8 +57,7 @@ tox -e py
 
 ### Integration tests
 
-Integration tests require the addition of [docker](https://docs.docker.com/engine/install/) or [podman](https://podman.io/getting-started/installation) and [docker-compose](https://docs.docker.com/compose/install/).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ This collection contains event source plugins, event filters and example ruleboo
 
 * ansible-rulebook >= 1.0.0
 * python >= 3.9
-* ansible >= 2.13
+* ansible-core >= 2.14
```
alinabuzachis, line 16: Maybe I am wrong, but ansible-core 2.14 has not reached EOL?

ssbarnea: Right, i will update this in a separate change.

alinabuzachis (dismissed):

alinabuzachis (approved):

PR #228, opened 2024-08-06
Ensure we use ubuntu 24.04 runners for testing

ssbarnea:
This should give us access to newer dependencies and avoid surprises
cause by use of ubuntu-latest mobile label.

Related: #219


ssbarnea:
To be noted that I checked that the failing integrations tests are the same before and after this change:
```
== 5 failed, 3 passed, 5 errors in 22.39s 
```

This change will allow us to install docker-compose on ubuntu using apt.

PR #230, opened 2024-08-06
Enable changelogs for collection

Akasurde:
##### SUMMARY

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,12 @@ repos:
         args: [--unique]
         # always_run: true
         files: ^(\.config/dictionary\.txt)$
+  - repo: https://github.com/ansible-community/antsibull-changelog
+    rev: 0.29.0
+    hooks:
+      - id: antsibull-changelog-lint
+      # Enable this when we have working changelog file
+      # - id: antsibull-changelog-lint-changelog-yaml
   - repo: https://github.com/koalaman/shellcheck-precommit
     rev: v0.10.0
     hooks:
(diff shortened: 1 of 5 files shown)


alinabuzachis (approved):

ssbarnea (changes requested):
While I find it useful, I am not seeing this being used by any GHA jobs, so I have no confidence that these settings are working.

Please ensure that we have a way to test them on CI.

Akasurde:
I think these are standard changelog settings. These will be used at the time of release. I tried them locally and they work.

ssbarnea (changes requested):
I started to work on it but I faced some problems like https://github.com/ansible-community/antsibull-changelog/issues/172 
and I had to stop.

We should not merge this unless we can validate that we can build/test these fragments.

changelogs/.plugin-cache.yaml
felixfontein, line 1: Don't check this file in. Add it to `.gitignore` instead.

.pre-commit-config.yaml
```diff
@@ -101,6 +101,11 @@ repos:
           - types-aiobotocore
           - watchdog
           - xxhash
+  - repo: https://github.com/ansible-community/antsibull-changelog
+    rev: 0.29.0
+    hooks:
+    - id: antsibull-changelog-lint
+    - id: antsibull-changelog-lint-changelog-yaml
```
felixfontein, line 108: If you lint `changelogs/changelog.yaml`, you also need to add that file. Otherwise linting it will fail.

PR #237, opened 2024-08-07
Update README to match the template

alinabuzachis:
This PR updates the README doc to match the [template](https://access.redhat.com/articles/7068606)


Akasurde (approved):

ssbarnea (changes requested):
TBH, I do not think that documentation should contain any "TBA" sections. Getting in sync with the template is not a bad thing, but not with empty TBA sections.

PS. I would be fine to have a partial change that is just missing these sections, and complete the rest later. I think that you could also comment them using `<!-- ... -->` should work to hide them but they will still be visible for those reading the non rendered readme file.

ssbarnea (approved):

PR #254, opened 2024-08-14
Add activation and activation_info modules

alinabuzachis:
Add activation and activation_info modules

### DIFF
--- a/tests/integration/targets/activation/tasks/main.yml
+++ b/tests/integration/targets/activation/tasks/main.yml
@@ -1,0 +1,268 @@
+---
+- block:
+    - set_fact:
+        credential_defaults: &credential_defaults
+          controller_username: "{{ controller_username }}"
+          controller_password: "{{ controller_password }}"
+          controller_host: "{{ controller_host }}"
+          validate_certs: false
+
+    - name: Generate a random_string for the test
+      set_fact:
+        random_string: "{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
+      when: random_string is not defined
+
+    - name: Generate a ID for the test
+      set_fact:
+        test_id: "{{ random_string | to_uuid }}"
+      when: test_id is not defined
+
+    - name: Define variables for credential and project
+      set_fact:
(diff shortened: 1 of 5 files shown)


ssbarnea (approved):
This LGTM but keep in mind that I do not have much experience with EDA itself.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -0,0 +1,269 @@
...
+        enabled: False
+        awx_token_name: "{{ awx_token_name }}"
+      register: _result
+
+    - name: Check rulebook activation creation
+      assert:
+        that:
+          - _result.changed
+
+    - name: Create a new rulebook activation again
```
matoval, line 170: I getting an error Create a new rulebook activation again
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to create/update rulebook activation: HTTP error {'detail': 'Method \"PATCH\" not allowed.'}"}`

alinabuzachis: I added a note about the fact that the module does not support PATCH and also handled that at the module level. Let me know if that makes sense to you. Thanks.

matoval: Makes sense, but I'm getting an error about the assert checking result message:

`[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}.
Found: 'A rulebook activation with name: {{ activation_name }} already exists. The module does not support
modifying a rulebook activation.' in _result.msg
fatal: [localhost]: FAILED! => {"msg": "The conditional check ''A rulebook activation with name: {{ activation_name }} already exists. The module does not support modifying a rulebook activation.' in _result.msg' failed. The error was: Conditional is marked as unsafe, and cannot be evaluated."}`

alinabuzachis: Removed jinja templating from the assert.

matoval (approved):
Ran activation integration tests and all ran successfully

PR #257, opened 2024-08-15
Add user module

Akasurde:
Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


plugins/modules/user.py
```diff
@@ -0,0 +1,224 @@
...
+      description:
+        - Set of roles to be associated with the user
+      choices: ["Admin", "Editor", "Contributor", "Operator", "Auditor", "Viewer"]
+      type: list
+      elements: str
+    update_secrets:
+      description:
+        - C(true) will always change password if user specifies password,
+          even if API gives $encrypted$ for password.
+        - C(false) will only set the password if other values change too.
```
alinabuzachis, line 59:
```suggestion
        - V(true) will always change password if user specifies password,
          even if API gives $encrypted$ for password.
        - V(false) will only set the password if other values change too.
```

plugins/modules/user.py
```diff
@@ -0,0 +1,224 @@
...
+        try:
+            ret = controller.delete_if_needed(user_type, endpoint=user_endpoint)
+        except EDAError as eda_err:
+            module.fail_json(msg=str(eda_err))
+        module.exit_json(**ret)
+
+    role_id = []
+    if roles:
+        for role in roles:
+            role_identifier = controller.resolve_name_to_id("roles", role)
```
alinabuzachis, line 186: I would enclose in a try except block to catch the case when the role does not exist.

matoval (approved):
Looks good, user integration tests all passed

alinabuzachis (approved):

ssbarnea (approved):

Issue #258, opened 2024-08-15
Enable mypy and ensure we pass with strict mode


ssbarnea:
A progressive approach would be highly desirable here, as we also have to merge a lot of new code into the repository.

- [ ] enable mypy via pre-commit with minimal mode
- [ ] address mypy exceptions made on 1st stage
- [ ] enable mypy strict mode

PR #264, opened 2024-08-15
Address more typing errors

ssbarnea:
Related: #258

### DIFF
--- a/extensions/eda/plugins/event_source/aws_cloudtrail.py
+++ b/extensions/eda/plugins/event_source/aws_cloudtrail.py
@@ -36,8 +36,8 @@ import json
 from datetime import datetime
 from typing import Any
 
-from aiobotocore.client import BaseClient
 from aiobotocore.session import get_session
+from botocore.client import BaseClient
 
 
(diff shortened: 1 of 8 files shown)


extensions/eda/plugins/event_source/aws_cloudtrail.py
```diff
@@ -36,8 +36,8 @@
 from datetime import datetime
 from typing import Any
 
-from aiobotocore.client import BaseClient
 from aiobotocore.session import get_session
+from botocore.client import BaseClient
```
Akasurde, line 40: This seems wrong to me. Changing `botocore` -> `aiobotocore` will change functionality.

Akasurde (changes requested):

extensions/eda/plugins/event_source/aws_cloudtrail.py
ssbarnea: I doubt it, see https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/client.py#L3 

Because they do not exported explicitly via __all_, that import is considered not work, so importing it from original location is the right approach.

PR #270, opened 2024-08-17
Add some critical types to module_utils and fix logic

ssbarnea:
This change adds minimal types and uncovers several logic bugs, where users of controller.get_one_or_many assumed it will not return a list.

I had to replace some calls of `get_one_or_many` with `get_exactly_one`, so please double check them.

Depends-On: #271

### DIFF
--- a/plugins/module_utils/controller.py
+++ b/plugins/module_utils/controller.py
@@ -2,7 +2,13 @@
 
 # Copyright: Contributors to the Ansible project
 # Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
-from __future__ import absolute_import, division, print_function
+from __future__ import absolute_import, annotations, division, print_function
+
+from typing import List, Optional, Union
+
+from ansible.module_utils.basic import AnsibleModule
+
+from .client import Client
 
 __metaclass__ = type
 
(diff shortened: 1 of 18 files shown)


plugins/module_utils/controller.py
```diff
@@ -81,8 +87,13 @@ def fail_wanted_one(self, response, endpoint, query_params):
         msg = f"Request to {display_endpoint} returned {response.json['count']} items, expected 1"
         raise EDAError(msg)
 
-    def get_exactly_one(self, endpoint, name=None, **kwargs):
-        return self.get_one_or_many(endpoint, name=name, allow_none=False, **kwargs)
+    def get_exactly_one(
+        self, endpoint, name=None, **kwargs
+    ) -> Optional[dict[str, bool]]:
+        result = self.get_one_or_many(endpoint, name=name, allow_none=False, **kwargs)
+        if isinstance(result, list):
```
alinabuzachis, line 94: This case is handled here https://github.com/ansible/event-driven-ansible/pull/270/files#diff-4fd0f63c1f7e4adaee865c4b96a5b4bdae623dad664c81e1f5a777823342d829R137. So, if you're handling this corner case in this function I would suggest to clean up get_one_or_many() at this point.

I wouldn't raise a RuntimeError, but I would suggest using something similar self.fail_wanted_one(response, endpoint, new_kwargs.get("data")) or raise an EDAError with a more descriptive message for the user.

ssbarnea: I switched to fail_wanted_one.

alinabuzachis: I you don't clean up the logic inside get_one_or_many() that handles the case when more than one resources are found, I do not think that your change will make any difference.

alinabuzachis (approved):

Akasurde (approved):

PR #274, opened 2024-08-19
Run integration tests in CI

alinabuzachis:
Run integration tests in CI


Akasurde (approved):

ssbarnea (approved):

ssbarnea:
To be noted that I will rework the workflow to integrate it inside existing one. Still, that is not a good enough reason to delay the merge.

PR #275, opened 2024-08-19
Cleanup module_utils and ensure consistence in the code base

alinabuzachis:
Cleanup module_utils and ensure consistence in the code base

Will rebase after https://github.com/ansible/event-driven-ansible/pull/274 is merged

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -5,16 +5,30 @@
 
 from __future__ import absolute_import, division, print_function
 
-from typing import Any, Optional, Union
-
 __metaclass__ = type
 
 
-def to_list_of_dict(
-    result: Optional[Union[dict[str, bool], list[dict[Any, Any]]]]
-) -> list[dict]:
-    if result is None:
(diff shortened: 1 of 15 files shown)


plugins/module_utils/common.py
```diff
@@ -5,16 +5,30 @@
...
+    params: Optional[dict[str, Any]] = None,
+):
+    result = None
+
+    try:
+        result = controller.resolve_name_to_id(
+            endpoint, name, **params if params is not None else {}
+        )
+    except EDAError as e:
+        module.fail_json(msg=f"Failed to lookup resource: {e}")
```
Akasurde, line 33: Do you think we should handle all `exit_json` and `fail_json` in the module code rather than module_utils code?

alinabuzachis: I added this utility function just because most of our modules were duplicating the same code. But if you think we shouldn't use it, I can remove it.

ssbarnea (approved):

PR #281, opened 2024-08-23
Add support for AAP 2.4

Akasurde:


plugins/modules/activation.py
```diff
@@ -341,7 +366,7 @@ def main() -> None:
         )
 
     # Activation Data that will be sent for create/update
-    activation_params = create_params(module, controller)
+    activation_params = create_params(module, controller, is_aap_24=is_aap_24)
```
ssbarnea, line 369: That sounds not future proof, how about passing a version instead of a boolean.

Akasurde: AFAIK, there is no EDA api which returns the version of AAP.

alinabuzachis (approved):

PR #295, opened 2024-08-30
Tune requirements filenames to meet dependabot behavior

ssbarnea:
As dependabot will gather `*.in` files from the folder as input for pip-compile and update all `*.txt` folders, we need to shuffle the filenames a little bit to keep it working
correctly.

If this does not happen, it will still attempt to update the `*.txt` file but without respecting the constraints imposed by the other input files.

Needed-By: #292
Related: https://github.com/dependabot/dependabot-core/issues/10007


alinabuzachis (approved):
Can you also please check and update the README.md file, especially when requirements are to be installed (if these changes make any changes to the process)?

PR #316, opened 2024-09-10
activation - rename webhooks option to event_streams and add rulebook_info module

alinabuzachis:
activation - rename webhooks option to event_streams and add rulebook_info module


Alex-Izquierdo:
Hi @alinabuzachis after deeper investigation and due to recent changes in the API REST, this PR as it is would not solve the issue, we need more things. I've started an internal thread to discuss them before decide how to solve it.

plugins/modules/activation.py
```diff
@@ -261,15 +280,43 @@ def create_params(
...
+                        msg=f"The specified source_index {event['source_index']} is out of range: {e}"
+                    )
+                event.pop("source_index")
+            elif event.get("source_name") and not any(
+                d["name"] == event.get("source_name") for d in sources
+            ):
+                module.fail_json(
+                    msg=f"The specified source_name {event.get('source_name')} does not exist."
+                )
+            activation_params["event_streams"].append(event)
```
Alex-Izquierdo, line 319: The param for the api is not "event_streams" but "source_mappings". 
This is to align the name with the UI. 
The payload for source_mapping is a yaml string that must meet the following schema:
```
  - source_name: 
    event_stream_id: 
    event_stream_name:
    rulebook_hash:
```

So, as it is now, what we have to do is: fetch as well the event_stream_id, and also the rulebook_hash for each event_stream and convert the list in a YAML to be passed to the "source_mappings" param.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -189,6 +197,109 @@
...
+        decision_environment_name: "{{ decision_env_name }}"
+        enabled: False
+        awx_token_name: "{{ awx_token_name }}"
+        organization_name: Default
+        event_streams:
+          - event_stream: "{{ event_stream_name }}"
+            source_index: 0
+      register: _result
+
+    - name: Check rulebook activation creation
```
Alex-Izquierdo, line 273: Would be good if we can retrieve the activation and ensure it is mapped to the event stream, as it is now that should fail because the event stream was never mapped since the api would ignore the wrong "event_streams" parameter sent at the creation.

plugins/modules/activation.py
alinabuzachis: I tried to apply those suggestions. Should I also rename the module's option to source_mappings or event_streams is ok?

plugins/modules/activation.py
```diff
@@ -107,11 +100,26 @@
...
-    elements: int
+    elements: dict
+    suboptions:
+      event_stream:
+        description:
+          - The name of the event stream.
+        type: str
+      source_name:
+        description:
+          - The name of the source.
```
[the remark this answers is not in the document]
alinabuzachis: Done!

plugins/modules/activation.py
Alex-Izquierdo: I think event_streams is ok because is the name in the UI. Nobody else has raised concerns about it. This is a mismatch between the UI and the API, so the module is in the middle. I don't have a strong opinion about but I think in the doubt, pairing it with the UI makes more sense. Do you agree?

alinabuzachis: Makes sense.

Alex-Izquierdo (approved):

jneedle (approved):

PR #319, opened 2024-09-12
Ensure that the mandatory fields required by the EDA API are respected within the modules.

alinabuzachis:
Ensure that the mandatory fields required by the EDA API are respected within the modules.

### DIFF
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -459,7 +459,7 @@ def main() -> None:
         (
             "state",
             "present",
-            ("name", "rulebook_name", "decision_environment_name"),
+            ("name", "rulebook_name", "decision_environment_name", "project_name"),
         )
     ]
 
--- a/plugins/modules/rulebook_info.py
+++ b/plugins/modules/rulebook_info.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
-DOCUMENTATION = """
+DOCUMENTATION = r"""
 ---
 module: rulebook_info
 author:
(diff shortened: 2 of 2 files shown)


plugins/modules/activation.py
```diff
@@ -9,7 +9,7 @@
 __metaclass__ = type
 
 
-DOCUMENTATION = """
+DOCUMENTATION = r"""
```
Akasurde, line 12: How about creating a separate PR to address the `r"""` changes?

Akasurde (dismissed):

Alex-Izquierdo:
Hi, @alinabuzachis I think the PR is incomplete unless your plan is to address it different PR's. According to the report:

- Every module that has an organization name, it must be mandatory. 

- Activation module: 
    * Decision_environment and rulebook_name are mandatory. 
    * Project_id is not mandatory in the api since it comes from the rulebook id, but project_name  must be mandatory in the module because the module needs it to find the rulebook id, otherwise it would fail if different projects has a rulebook with the same name. 

- Credential Type:
    * inputs and injectors are mandatory

- Event streams:
    * event stream type and credential name are mandatory

- Projects: 
    * url is mandatory

alinabuzachis:
> * Decision_environment

I just rebased the PR and this seems to be the only thing missing. From what I can see, all the other requirements are already met.
For modules that require an `organization_name`, even if the `organization_name` is not added directly to required_if, it is required through this type of check (e.g., https://github.com/ansible/event-driven-ansible/blob/main/plugins/modules/project.py#L139)

Alex-Izquierdo (approved):

Akasurde (approved):

ssbarnea (dismissed):

ssbarnea (approved):

Alex-Izquierdo (approved):

PR #320, opened 2024-09-12
Ensure modules are idempotent

alinabuzachis:
Ensure modules are idempotent

### DIFF
--- a/plugins/module_utils/controller.py
+++ b/plugins/module_utils/controller.py
@@ -242,6 +242,7 @@ class Controller:
             return True  # all sub-fields are either equal or could be equal
         if old_field == Controller.ENCRYPTED_STRING:
             return True
+
         return bool(new_field == old_field)
 
     def objects_could_be_different(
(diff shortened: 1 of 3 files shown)


Akasurde (commented):
Do you think we need a changelog for this? Rest LGTM

Alex-Izquierdo:
> Do you think we need a changelog for this? Rest LGTM

@Akasurde how the change log works? I thought it would be generated automatically at the release time. If we need to do something on purpose I think https://github.com/ansible/event-driven-ansible/pull/322 should have also to update the change log.

Akasurde:
@Alex-Izquierdo 
See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example for comment. These fragments will create changelog at the time of release.

alinabuzachis:
> @Alex-Izquierdo See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example for comment. These fragments will create changelog at the time of release.

The problem is that I do not see anymore the changelogs folder.

Akasurde:
> > @Alex-Izquierdo See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example for comment. These fragments will create changelog at the time of release.
> 
> The problem is that I do not see anymore the changelogs folder @Akasurde .


Seems like GitHub release is managing the changelog. NVM, drop this review request comment.

Akasurde (approved):

Alex-Izquierdo (approved):

PR #321, opened 2024-09-12
event_stream - add support for headers and forward_events options

alinabuzachis:
event_stream - add support for `headers` and `forward_events` options

### DIFF
--- a/plugins/modules/event_stream.py
+++ b/plugins/modules/event_stream.py
@@ -51,6 +51,19 @@ options:
     type: str
     aliases:
       - type
+  headers:
+    description:
+      - Comma separated HTTP header keys that you want to include in the event payload.
+        To include all headers in the event payload, leave the field empty.
+    type: str
+    default: ''
+    version_added: 2.1.0
+  forward_events:
+    description:
+      - Enable the event stream to forward events to the rulebook activation where it is configured.
+    type: bool
+    default: True
+    version_added: 2.1.0
   state:
     description:
       - Desired state of the resource.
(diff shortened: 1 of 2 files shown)


plugins/modules/event_stream.py
```diff
@@ -51,6 +51,19 @@
     type: str
     aliases:
       - type
+  headers:
+    description:
+      - Comma separated HTTP header keys that you want to include in the event payload.
```
Akasurde, line 56: Does it make sense to have a dictionary instead of a comma-separated string? Not a blocker, just curious to know.

Akasurde: A dictionary would be cleaner to write, internally we can always transform it into a comma-separated string. Not a blocker.

Alex-Izquierdo: what should be the key and values of that content? It would make more sense a list instead.

Akasurde: something similar to - 

```
    headers:
      Content-Type: application/json
      x-ms-client-workspace-path: /subscriptions/{{ sub_id }}/resourcegroups/{{ res_group }}/providers/microsoft.operationalinsights/workspaces/{{ w_spaces }}
      x-ms-client-platform: ibiza
      x-ms-client-auth-token: "{{ token_az }}"
```

Alex-Izquierdo: I think you may be misunderstanding the purpose of that parameter. It is a list of headers that come in the POST request made to the Event Stream endpoint that are included in the event for ansible-rulebook. 
A dict doesn't make sense because the value comes from the request. It should be a list of headers anyway, which is what the API expects.

Akasurde: OK. Make sense.

Alex-Izquierdo (dismissed):

Akasurde (dismissed):

Alex-Izquierdo (approved):

Akasurde (approved):

PR #332, opened 2024-10-02
feat: add sync capability to project module

Alex-Izquierdo:
**Why:**
Project module misses the sync capability. An existing project can be synced (pull changes from the repo) and this have be able to be done from the module. 
jira: https://issues.redhat.com/browse/AAP-32264

**How:**
We remove the constraint to define the url when state is present because the project might already exists, url is only required when the project needs to be created, we check it in advance to check the requirement of the url and to check if the sync operation can be performed. We add the new parameter "sync" to run the synchronization (it is an async task in the backend) if the project already exists. The client needs to be updated because the backend can also return 202 for a POST operation. 

Include:
- A new sync parameter and logic related. 
- A new generic "create" method to run a post request
- A fix of `get_exactly_one` helper that generated undesired behaviors when a name search returns results based on a partial match. 
- A minor fix by initializing `project_type` variable which could be unbound. 
- Tests added and updated for the new behavior.
- A minor fix in tests that they had defined unnecessary variables that prevented to execute the tests locally. 
- Updates the project url in tests with a public valid url.

### DIFF
--- a/plugins/module_utils/controller.py
+++ b/plugins/module_utils/controller.py
@@ -83,25 +83,23 @@ class Controller:
         raise EDAError(msg)
 
     def get_exactly_one(
-        self, endpoint: str, name: Optional[str] = None, **kwargs: Any
+        self, endpoint: str, name: Optional[str], **kwargs: Any
     ) -> dict[str, Any]:
         result = self.get_one_or_many(endpoint, name=name, **kwargs)
-        if len(result) == 0:
-            return {}
-        if len(result) > 1:
-            if name:
-                # Since we did a name or ID search and got > 1 return
-                # something if the id matches
-                for asset in result:
-                    if str(asset["id"]) == name:
-                        return asset
-            # We got > 1 and either didn't find something by ID (which means
-            # multiple names)
-            # Or we weren't running with a or search and just got back too
-            # many to begin with.
-            self.fail_wanted_one(result)
+        matches = []
+        name_field = self.get_name_field_from_endpoint(endpoint)
 
-        return result[0]
+        for asset in result:
+            if str(asset[name_field]) == name:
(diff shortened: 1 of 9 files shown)


plugins/module_utils/controller.py
```diff
@@ -184,6 +184,23 @@ def create_if_needed(
...
+        response = self.post_endpoint(endpoint, data=new_item)
+
+        if response.status in [201, 202]:
+            self.result["changed"] = True
+            return self.result
+
+        if response.json:
+            error_msg = f"Unable to create {item_type} {new_item}: {response.json}"
+
+        error_msg = f"Unable to create {item_type} {new_item}: {response.status} {response.data}"
```
jshimkus-rh, line 201: ¿If there is a json decodable response set error_msg but then replace it with a different error_msg regardless?

Alex-Izquierdo: good catch, thanks. Fixed

jshimkus-rh (approved):

Akasurde (dismissed):

alinabuzachis (approved):

jshimkus-rh (approved):

zkayyali812:
The environments/deployments here are new to me. Should these be passing now? 

Once the tox checks pass I will approve this PR if those are broken

Alex-Izquierdo:
@zkayyali812 @jshimkus-rh This is the good. 
Still failing a check for coverage, I'm looking how to skip it because it is already covered by integration tests. 

@Akasurde @alinabuzachis Review again if you can please. The main change is the get_exactly_one, I have changed its logic because the current implementation didn't do what it was expected. I basically changed it to support partial matches.

zkayyali812 (approved):
/lgtm!

PR #336, opened 2024-10-08
Fix integration tests

alinabuzachis:
Fix integration tests by ensuring the appropriate event_stream_type is used

### DIFF
--- a/tests/integration/targets/activation/tasks/main.yml
+++ b/tests/integration/targets/activation/tasks/main.yml
@@ -225,7 +225,7 @@
         name: "{{ event_stream_name }}"
         credential_name: "{{ credential_name_basic }}"
         organization_name: Default
-        event_stream_type: Basic Event Stream
+        event_stream_type: "basic"
       register: _result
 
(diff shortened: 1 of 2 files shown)


tests/integration/targets/activation/tasks/main.yml
```diff
@@ -225,7 +225,7 @@
         name: "{{ event_stream_name }}"
         credential_name: "{{ credential_name_basic }}"
         organization_name: Default
-        event_stream_type: Basic Event Stream
+        event_stream_type: "basic"
```
Alex-Izquierdo, line 228: I'm not sure what are you trying to fix, but the right type is `Basic Event Stream`

Alex-Izquierdo: Ok, sorry, I mislead it with the credential type

Alex-Izquierdo:
https://github.com/ansible/event-driven-ansible/pull/332 this has the fix for the tox/check error

Alex-Izquierdo (approved):

PR #337, opened 2024-10-08
event_stream - Ensure forward_events works as expected

alinabuzachis:
event_stream - Ensure forward_events works as expected

### DIFF
--- a/plugins/modules/event_stream.py
+++ b/plugins/modules/event_stream.py
@@ -62,7 +62,7 @@ options:
     description:
       - Enable the event stream to forward events to the rulebook activation where it is configured.
     type: bool
-    default: true
+    default: false
     version_added: 2.1.0
   state:
     description:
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (commented):
Hi @alinabuzachis , the option name is correct, because "forward events" is what we have right now in the UI. This is a similar case that we had with source mappings, the UI has a different interface than the API, so we decided the module to align with the UI. 

In this way, we don't need to rename or add new fields, just we have to reverse the boolean for the test_mode api param. 
test_mode = not forward_events

plugins/modules/event_stream.py
```diff
@@ -62,7 +62,7 @@
     description:
       - Enable the event stream to forward events to the rulebook activation where it is configured.
     type: bool
-    default: True
+    default: true
```
Alex-Izquierdo, line 65: We should change the default value to be aligned with UI and API

zkayyali812 (approved):

Issue #130, opened 2023-05-22
file_watch raises QueueFull exception on event trigger


rcarrillocruz:
I have the following rulebook:

```
- name: Local Test Runner Rules
  hosts: all
  sources:
    - name: file_watch
      ansible.eda.file_watch:
        path: "/home/cloud-user/folder-to-watch"
        recursive: true
  rules:
    - name: run tests 1
      condition: event.type == "DirModifiedEvent"
      action:
        debug:
```

Then, I run it with rulebook with:

```
[cloud-user@ricky-aap-containerized-installer eda]$ ansible-rulebook -i inventory.yml --rulebook rulebook.yml --verbose
2023-05-22 15:57:29,369 - ansible_rulebook.app - INFO - Starting sources
2023-05-22 15:57:29,369 - ansible_rulebook.app - INFO - Starting rules
2023-05-22 15:57:29,369 - ansible_rulebook.engine - INFO - run_ruleset
2023-05-22 15:57:29,370 - drools.ruleset - INFO - Using jar: /home/cloud-user/.local/lib/python3.9/site-packages/drools/jars/drools-ansible-rulebook-integration-runtime-1.0.1-SNAPSHOT.jar
2023-05-22 15:57:29 987 [main] INFO org.drools.ansible.rulebook.integration.api.rulesengine.AbstractRulesEvaluator - Start automatic pseudo clock with a tick every 100 milliseconds
2023-05-22 15:57:29,991 - ansible_rulebook.engine - INFO - ruleset define: {"name": "Local Test Runner Rules", "hosts": ["all"], "sources": [{"EventSource": {"name": "file_watch", "source_name": "ansible.eda.file_watch", "source_args": {"path": "/home/cloud-user/folder-to-watch", "recursive": true}, "source_filters": []}}], "rules": [{"Rule": {"name": "run tests 1", "condition": {"AllCondition": [{"EqualsExpression": {"lhs": {"Event": "type"}, "rhs": {"String": "DirModifiedEvent"}}}]}, "actions": [{"Action": {"action": "debug", "action_args": {}}}], "enabled": true}}]}
2023-05-22 15:57:30,003 - ansible_rulebook.engine - INFO - load source
2023-05-22 15:57:30,615 - ansible_rulebook.engine - INFO - load source filters
2023-05-22 15:57:30,615 - ansible_rulebook.engine - INFO - loading eda.builtin.insert_meta_info
[...]

PR #342, opened 2024-10-21
fix: Avoid QueueFull exception in file_watch source plugin

Alex-Izquierdo:
Avoid QueueFull exception by using queue.put instead of queue.put_nowait, no need to catch the exception, await queue.put() implements backpressure natively.

Closes https://github.com/ansible/event-driven-ansible/issues/130

### DIFF
--- a/extensions/eda/plugins/event_source/file_watch.py
+++ b/extensions/eda/plugins/event_source/file_watch.py
@@ -41,47 +41,55 @@ def watch(
             RegexMatchingEventHandler.__init__(self, **kwargs)
 
         def on_created(self: "Handler", event: FileSystemEvent) -> None:
-            loop.call_soon_threadsafe(
-                queue.put_nowait,
-                {
-                    "change": "created",
-                    "src_path": event.src_path,
-                    "type": event.__class__.__name__,
-                    "root_path": root_path,
-                },
+            asyncio.run_coroutine_threadsafe(
+                queue.put(
+                    {
+                        "change": "created",
+                        "src_path": event.src_path,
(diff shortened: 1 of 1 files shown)


extensions/eda/plugins/event_source/file_watch.py
```diff
@@ -41,47 +41,55 @@ def __init__(self: "Handler", **kwargs: Any) -> None:
...
         def on_created(self: "Handler", event: FileSystemEvent) -> None:
-            loop.call_soon_threadsafe(
-                queue.put_nowait,
-                {
-                    "change": "created",
-                    "src_path": event.src_path,
-                    "type": event.__class__.__name__,
-                    "root_path": root_path,
-                },
+            asyncio.run_coroutine_threadsafe(
```
mkanoor, line 44: @Alex-Izquierdo Should this be using the await put call like the other source [plugins](https://github.com/ansible/event-driven-ansible/blob/586ea78462ebe506b788224d346011bea3c8f3b6/extensions/eda/plugins/event_source/aws_sqs_queue.py#L82).

Alex-Izquierdo: @mkanoor if you look at the code, watchdog is sync code that is executed with a [threadpool](https://github.com/ansible/event-driven-ansible/blob/fd3e8e1695acaedac9a5c8a88fbdf8604ab02e60/extensions/eda/plugins/event_source/file_watch.py#L111) this is not a pure asyncio plugin.

mkanoor (approved):

zkayyali812 (approved):

PR #362, opened 2024-11-25
Upgrade tox-extra to 2.1.0

ssbarnea:


kaiokmo:
@ssbarnea this one's stale for quite some time now. Do you still think you want to proceed with this change?

Issue #365, opened 2024-11-27
Cannot create Rulebook Activations as disabled 


zjleblanc:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/ansible-rulebook/issues) for duplicates.
- [X] I understand that event-driven-ansible collection is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Using the rulebook_activation module, one cannot create resource in the disabled state. It will always be created as enabled.

### Environment

[de-support-rhel9:1.1.1-31](https://catalog.redhat.com/software/containers/ansible-automation-platform-25/de-supported-rhel9/650a5674ad524b664b693729)

ansible-rulebook-1.1.1-1.el9ap.noarch

### Steps to reproduce

Define a rulebook_activiation in configuration-as-code variables for use with infra.aap_configuration
```
---
eda_rulebook_activations:
  - name: Demo Rulebook Activation
    description: Rulebook that processes various events sourced from Dynatrace problems
    organization: Default
    project: EDA Demos
    rulebook: dynatrace_event_stream.yml
    decision_environment: Default Decision Environment
    event_streams: 
      - event_stream: Dynatrace Event Stream
        source_name: Dynatrace Event Source
    eda_credentials:
      - My Controller
    enabled: false
```
Ensure the variables file is loaded
Run a playbook and invoke `infra.aap_configuration.dispatch`

### Actual results

Rulebook activation **Demo Rulebook Activation** is created in the enabled state

### Expected results

Rulebook activation **Demo Rulebook Activation** is created in the disabled state

leogallego:
Hey Zack,  this was fixed with your PR, not sure why it didn't close automatically. Pinging so others don't fall in the trap of thinking this is still open. 

https://github.com/ansible/event-driven-ansible/pull/366

PR #366, opened 2024-11-27
Check if param flag values are defined, not truthiness

zjleblanc:
Addresses #365 by checking if module parameters which represent flags are defined, preventing evaluation of their truthyness. Current behavior will ignore an incoming value of false and default back to true because of the conditional logic:

Example:
`module.params["enabled"] = False`
`if module.params.get("enabled")` evaluates to False, but the intent is to determine if it has been defined, so we want this to evaluate to True.

This PR addresses the same issue for `swap_single_source` param.

### DIFF
--- a/tests/integration/targets/activation/tasks/main.yml
+++ b/tests/integration/targets/activation/tasks/main.yml
@@ -273,7 +273,7 @@
 
     - name: Assert that rulebook activation is disabled
       ansible.builtin.assert:
-        that: 
+        that:
           - not _result_activation_info.activations[0].is_enabled
 
     - name: List all the rulebook activations
@@ -413,7 +413,7 @@
 
     - name: Assert that swap single source flag is disabled
       ansible.builtin.assert:
-        that: 
+        that:
           - not _result_activation_info.activations[0].swap_single_source
 
     - name: Delete project
--- a/tests/integration/targets/activation/tasks/main.yml
+++ b/tests/integration/targets/activation/tasks/main.yml
@@ -269,6 +269,11 @@
     - name: Get information about the rulebook activation
       ansible.eda.rulebook_activation_info:
         name: "{{ activation_name }}"
+      register: _result_activation_info
+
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
Hi @zjleblanc 
Sorry for the late response, I was on holidays. 
I've checking it deeply and your PR already addresses all the issues, we don't need more changes. The only thing is that we should test this specific issue. Would you mind to add a new task in https://github.com/ansible/event-driven-ansible/blob/main/tests/integration/targets/activation/tasks/main.yml so we can verify that `enabled` param is correctly created? 

Otherwise I can recreate the PR with the test. 

Thank you very much.

zjleblanc:
@Alex-Izquierdo added an assert based on the existing integration tasks - will check back to make sure it passes.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -404,6 +411,11 @@
         that:
           - event_stream_name in (source_mappings_dict | map(attribute='event_stream_name') | list)
 
+    - name: Assert that swap single source flag is disabled
+      ansible.builtin.assert:
+        that:
+          - not _result_activation_info.activations[0].swap_single_source
```
Alex-Izquierdo, line 417: We don't need this assertion, it is a feature of the collection for configuring event streams. not a field in the backend

Alex-Izquierdo, line 417: We don't need this assertion, it is a feature of the collection for configuring event streams. not a field in the backend

Alex-Izquierdo (changes requested):

tests/integration/targets/activation/tasks/main.yml
zjleblanc: got it, removed this assertion 👍

Alex-Izquierdo (approved):
LGTM, thank you very much for your contribution @zjleblanc

Dostonbek1 (approved):

PR #384, opened 2025-01-27
fix: AAP-39412 incorrect credential in examples

kaiokmo:
https://issues.redhat.com/browse/AAP-39412

The eda-setup.yml example playbook currently fails because the credential name passed to create a basic event stream is different from the credential created before.

This commit fixes this issue and allows the example playbook to continue.

### DIFF
--- a/examples/eda-setup.yml
+++ b/examples/eda-setup.yml
@@ -80,7 +80,7 @@
       ansible.eda.event_stream:
         organization_name: "{{ organization_name }}"
         name: basic event stream
-        credential_name: basic event stream
+        credential_name: basic event stream credential
         event_stream_type: Basic Event Stream
 
     - name: Get the event stream generated


Alex-Izquierdo (dismissed):

examples/eda-setup.yml
```diff
@@ -80,7 +80,7 @@
       ansible.eda.event_stream:
         organization_name: "{{ organization_name }}"
         name: basic event stream
-        credential_name: basic event stream
+        credential_name: basic event stream credential
         event_stream_type: Basic Event Stream
```
Alex-Izquierdo, line 84: Can you also remove this line? It is not necessary and can confuse users.

kaiokmo: done

Alex-Izquierdo (approved):

alinabuzachis (approved):

PR #385, opened 2025-01-28
feat: [AAP-38755] add sidecar documentation files for EDA plugins

Dostonbek1:
This PR adds sidecar documentation files for both event source and event filters plugins, by following the format outlined [here](https://docs.ansible.com/ansible-core/devel/dev_guide/sidecar.html).


Alex-Izquierdo:
There is no reason to add the documentation in separated files. According to the documentation:
```
YAML documentation is mainly intended for filters, tests and modules. While it is possible to use with other plugin types, Ansible always recommends having documentation in the same file as the code for most cases.
```

These plugins are not written in other language than python and by design the module can not contain multiple plugins. And I personally feel better to leave it in the same file if there is no a good reason for it. The regular modules also have their documentation in the same file, so we can keep this consistency.

extensions/eda/plugins/event_source/pg_listener.py
```diff
@@ -81,6 +24,53 @@
...
 
+DOCUMENTATION = r"""
+---
+author:
+  - Doston Toirov (@dtoirov)
+short_description: Read events from pg_pub_sub.
+description:
+  - An ansible-rulebook event source plugin for reading events from pg_pub_sub.
+version_added: '2.4.0'
+options:
```
Alex-Izquierdo, line 35: This is wrong. 
Look at the original: https://github.com/ansible/event-driven-ansible/pull/385/files#diff-9044cb1d0c37a42fa4acb893834fdd6be6edc60ce1ded948929e6f78806413abL12

Dostonbek1: corrected.

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -1,47 +1,68 @@
...
+    type: bool
+    default: "false"
+  log_error:
+    description:
+      - Whether log an error message if host_path does not
+        exist in the event.
+      - You can turn if off if it is expected to have events not
+        having the host_path to avoid noises in the log.
+    type: bool
+    default: "bool"
```
Alex-Izquierdo, line 52: if the type is bool, the default value can not be a string

extensions/eda/plugins/event_filter/json_filter.py
```diff
@@ -1,23 +1,29 @@
...
+version_added: '2.4.0'
+options:
+  exclude_keys:
+    description:
+      - A list of strings or patterns to remove.
+    type: list
+  include_keys:
+    description:
+      - A list of strings or patterns to keep even if it matches exclude_keys patterns.
+    type: list
```
Alex-Izquierdo, line 24: missing the additional "elements: str"

extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -1,47 +1,68 @@
...
+DOCUMENTATION = r"""
+---
+author:
+  - Doston Toirov (@dtoirov)
+short_description: Extract hosts from the event data and insert them to the meta dict.
+description:
+  - An ansible-rulebook event filter that extracts hosts from the event data and
+    inserts them to the meta dict. Ansible-rulebook will limit an ansible action
+    running on hosts in the meta dict.
+version_added: '2.4.0'
```
Alex-Izquierdo, line 17: Why this field in all the plugins? All of these options/plugins were not added in 2.4.0, We can just ignore it in all of them. This field would be used for every new param added from now.

extensions/eda/plugins/event_filter/normalize_keys.py
```diff
@@ -1,47 +1,48 @@
...
+  - If there is a existing key with the normalized name, it will get overwritten
+    by default. If you don't want to over write it you can pass in "overwrite: false"
+    The default value of overwrite is true.
+version_added: '2.4.0'
+options:
+  overwrite:
+    description:
+      - Overwrite the values if there is a collision with a new key.
+    type: bool
+    default: "true"
```
Alex-Izquierdo, line 28: default can not be a string, the type is bool

extensions/eda/plugins/event_filter/dashes_to_underscores.py
```diff
@@ -1,16 +1,25 @@
...
+  overwrite:
+    description:
+      - Overwrite the values if there is a collision with a new key.
+    type: bool
+    default: true
 """
 
-import multiprocessing as mp
-from typing import Any
+EXAMPLES = r"""
```
Alex-Izquierdo, line 19: mmm.... The example of the dashes_to_underscores filter does not use the filter

PR #387, opened 2025-01-31
feat [AAP-37306]: Add support to edit and copy activations

kaiokmo:
https://issues.redhat.com/browse/AAP-37306


kaiokmo:
Abandoned in favor of #394

PR #388, opened 2025-02-03
feat: [AAP-38755] add documentation for event filter plugins

Dostonbek1:
This PR adds new documentation format for event filters plugins as part of our efforts in supporting EDA plugin documentation. 

### DIFF
--- a/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
+++ b/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
@@ -1,38 +1,59 @@
-"""insert_hosts_to_meta.py.
+DOCUMENTATION = r"""
+---
+author:
+  - Doston Toirov (@dtoirov)
+short_description: Extract hosts from the event data and insert them to the meta dict.
+description:
+  - An ansible-rulebook event filter that extracts hosts from the event data and
+    inserts them to the meta dict. Ansible-rulebook will limit an ansible action
+    running on hosts in the meta dict.
+options:
+  host_path:
+    description:
+      - The json path inside the event data to find hosts.
+      - Do nothing if the key is not present or does exist in event.
+    type: str
+    default: null
+  path_separator:
+    description:
+      - The separator to interpret host_path.
+    type: str
+    default: "."
+  host_separator:
+    description:
+      - The separator to interpret host string.
+      - host_path can point to a string or a list. If it is a single
+        string but contains multiple hosts, use this parameter to
+        delimits the hosts. Treat the value as a single host if the
+        parameter is not present.
+    type: str
+    default: null
+  raise_error:
+    description:
+      - Whether raise PathNotExistError if host_path does not
+        exist in the event.
+      - It is recommended to turn it on during the rulebook
+        development time. You can then turn it off for production.
+    type: bool
+    default: false
+  log_error:
+    description:
+      - Whether log an error message if host_path does not
+        exist in the event.
+      - You can turn if off if it is expected to have events not
+        having the host_path to avoid noises in the log.
+    type: bool
+    default: true
+"""
 
-An ansible-rulebook event filter that extract hosts from the event data and
-insert them to the meta dict. Ansible-rulebook will limit an ansible action
-running on hosts in the meta dict.
-
-Arguments:
----------
-    host_path:      The json path inside the event data to find hosts.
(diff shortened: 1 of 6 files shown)


extensions/eda/plugins/event_filter/insert_hosts_to_meta.py
```diff
@@ -1,47 +1,68 @@
...
 import dpath
 
+DOCUMENTATION = r"""
+---
+author:
+  - Doston Toirov (@dtoirov)
+short_description: Extract hosts from the event data and insert them to the meta dict.
+description:
+  - An ansible-rulebook event filter that extracts hosts from the event data and
+    inserts them to the meta dict. Ansible-rulebook will limit an ansible action
```
mkanoor, line 15: @Dostonbek1 should Ansible-rulebook be ansible-rulebook or may be write it as "In ansible-rulebool this will limit an action" Its not anisble action.

extensions/eda/plugins/event_filter/json_filter.py
```diff
@@ -1,23 +1,39 @@
...
+options:
+  exclude_keys:
+    description:
+      - A list of strings or patterns to remove.
+    type: list
+    elements: str
+    default: null
+  include_keys:
+    description:
+      - A list of strings or patterns to keep even if it matches exclude_keys patterns.
```
mkanoor, line 24: @Dostonbek1 for botth the exclude_keys and include_keys we have to mention that it is a path as shown here https://github.com/mkanoor/eda-project/blob/e4a8170c8fb2565e4633673694c4bb83b4c1a8e0/rulebooks/webhook_with_filter.yml#L20 so if the data is under key1.key2.key3 it would have to be listed as 
- key1
- key2
- key3
Since they are nested path, we can't just exclude key3 it needs to be the full path.

Another example https://github.com/ansible/ansible-rulebook/blob/b4f8b6c69be2d84f3e6fc6bd61cd0dab41b2ac43/tests/examples/60_json_filter.yml#L5

Dostonbek1: pushed the suggested changes.

extensions/eda/plugins/event_filter/noop.py
```diff
@@ -1,7 +1,14 @@
-"""noop.py:   An event filter that does nothing to the input."""
-
 from typing import Any
 
+DOCUMENTATION = r"""
+---
+author:
+  - Doston Toirov (@dtoirov)
+short_description: Do nothing.
+description:
```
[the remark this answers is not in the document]
Dostonbek1: I can put a separate PR to remove it, once we merge this one.

mkanoor (dismissed):

extensions/eda/plugins/event_filter/noop.py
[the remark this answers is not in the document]
Alex-Izquierdo: @mkanoor We can not remove plugins arbitrarily, that could be a breaking change if it's not correctly handled. We have pending to discuss the overall content of the collection, either to move plugins to the eventual community collection or just remove them completely.

extensions/eda/plugins/event_filter/normalize_keys.py
```diff
@@ -1,46 +1,46 @@
...
-import multiprocessing as mp
-import re
-from typing import Any
+EXAMPLES = r"""
+- sources:
+    - my_source:
+        ....
+      filters:
+        ansible.eda.normalize_keys:
+- name: prevent overwriting of keys
```
Alex-Izquierdo, line 36: This is confusing. The source has also a name field. Use a newline to distinguish different sources. Better if you remove the "sources" key root. That would be aligned to regular ansible modules, which have the task definition without the `tasks` key root.

Alex-Izquierdo (commented):
The examples are not consistent. Some of them only include the `filters` content, other times are the `sources` content, even one of them has a whole rulebook. 
If we follow the regular ansible modules, where the example only includes the task definition, since the filters belong to the sources definition, IMO the whole source is what we should put in the example. Similar to what you have in extensions/eda/plugins/event_filter/normalize_keys.py but without the key root `sources`

Example: https://github.com/ansible/event-driven-ansible/blob/main/plugins/modules/credential.py#L64

Alex-Izquierdo (approved):

mkanoor (approved):

PR #392, opened 2025-02-04
feat: [AAP-38755] add documentation for pg_listener, range, tick, url_check and webhook plugins

Dostonbek1:
This PR adds documentation for both event sources pg_listener, range, tick, url_check and webhook plugins, by following the format outlined [here](https://docs.ansible.com/ansible-core/devel/dev_guide/sidecar.html).


extensions/eda/plugins/event_source/pg_listener.py
```diff
@@ -63,6 +6,68 @@
...
+    elements: str
+    required: true
+notes:
+  - Chunking - this is just informational, a user doesn't have to do anything
+    special to enable chunking. The sender which is the pg_notify
+    action from ansible rulebook will decide if chunking needs to
+    happen based on the size of the payload.
+  - If the messages are over 7KB the sender will chunk the messages
+    into separate payloads with each payload having having the following
+    keys
```
ttuffin, line 41:
```suggestion
    special to enable chunking. The sender, which is the pg_notify
    action from ansible-rulebook, will decide if chunking needs to
    happen based on the size of the payload.
  - If the messages are over 7KB the sender will chunk the messages
    into separate payloads with each payload having the following
    keys:
```

Alex-Izquierdo (approved):

ttuffin (approved):

PR #394, opened 2025-02-12
feat [AAP-37306]: add support update and copy activation

kaiokmo:
https://issues.redhat.com/browse/AAP-37306

This implements the logic to edit/update and copy an activation, as well as enable and disable activations.

### DIFF
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -58,7 +58,6 @@ options:
   restart_policy:
     description:
       - The restart policy for the rulebook activation.
-    default: "on-failure"
     choices: ["on-failure", "always", "never"]
     type: str
   enabled:
@@ -141,7 +140,6 @@ options:
       - This parameter is supported in AAP 2.5 and onwards.
         If specified for AAP 2.4, value will be ignored.
     type: str
-    default: "error"
     choices: ["debug", "info", "error"]
   state:
     description:
@@ -164,8 +162,9 @@ EXAMPLES = r"""
     project_name: "Example Project"
     rulebook_name: "hello_controller.yml"
     decision_environment_name: "Example Decision Environment"
-    enabled: false
-    awx_token_name: "Example Token"
+    eda_credentials:
+        - My AAP Credentials
+    organization_name: "Default"
 
 - name: Create a rulebook activation with event_streams option
   ansible.eda.rulebook_activation:
@@ -174,12 +173,43 @@ EXAMPLES = r"""
     project_name: "Example Project"
     rulebook_name: "hello_controller.yml"
     decision_environment_name: "Example Decision Environment"
-    enabled: false
-    awx_token_name: "Example Token"
     organization_name: "Default"
+    eda_credentials:
+        - My AAP Credentials
     event_streams:
       - event_stream: "Example Event Stream"
         source_name: "Sample source"
+
+- name: Update rulebook activation fields
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation"
+    eda_credentials:
+        - My AAP credential
+        - My other credential
+    log_level: debug
+    restart_policy: always
+    project_name: "Example Project"
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
+
+- name: Copy an existing rulebook activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation - copy"
+    copy_from: "Example Rulebook Activation"
+    project_name: "Example Project"
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
+
+- name: Rename rulebook activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation - copy"
+    new_name: "Renamed Example Rulebook Activation"
+    project_name: "Example Project"
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
 
 - name: Delete a rulebook activation
   ansible.eda.rulebook_activation:
@@ -399,8 +429,6 @@ def create_params(
 
     if eda_credential_ids is not None:
         activation_params["eda_credentials"] = eda_credential_ids
-    else:
-        activation_params["eda_credentials"] = []
 
     if not is_aap_24 and module.params.get("k8s_service_name"):
         activation_params["k8s_service_name"] = module.params["k8s_service_name"]
@@ -471,7 +499,6 @@ def main() -> None:
         extra_vars=dict(type="str"),
         restart_policy=dict(
             type="str",
-            default="on-failure",
             choices=[
                 "on-failure",
                 "always",
@@ -504,7 +531,7 @@ def main() -> None:
             removed_in_version="3.0.0",
             removed_from_collection="ansible.eda",
         ),
-        log_level=dict(type="str", choices=["debug", "info", "error"], default="error"),
+        log_level=dict(type="str", choices=["debug", "info", "error"]),
         state=dict(
             choices=["present", "absent", "enabled", "disabled"], default="present"
         ),
@@ -583,8 +610,11 @@ def main() -> None:
                 copy_endpoint = f"activations/{activation["id"]}/copy"
                 params = {"name": name}
 
-                result = controller.post_endpoint(endpoint=copy_endpoint, data=params)
-                module.exit_json(**result)
+                controller.post_endpoint(
+                    endpoint=copy_endpoint,
+                    data=params
+                )
+                module.exit_json(changed=True)
             except EDAError as e:
                 module.fail_json(msg=f"Failed to copy rulebook activation: {e}")
 
@@ -597,9 +627,8 @@ def main() -> None:
         activation_params.pop("enabled", None)
 
         # Change from list of credentials to a list of IDs in existing activation
-        cred_ids = {cred_id["id"] for cred_id in activation["eda_credentials"]}
-        if set(activation_params["eda_credentials"]) == cred_ids:
-            activation["eda_credentials"] = activation_params["eda_credentials"]
+        credential_ids = [credential_id["id"] for credential_id in activation["eda_credentials"]]
+        activation["eda_credentials"] = credential_ids
 
         if controller.objects_could_be_different(activation, activation_params):
             try:
@@ -614,7 +643,9 @@ def main() -> None:
                 # while it's still enabled. We simply honor the operation first before
                 # anything else.
                 if op_type != NO_OP:
-                    enable_disable_endpoint = f"activations/{activation["id"]}/{op_type}"
+                    enable_disable_endpoint = (
+                        f"activations/{activation["id"]}/{op_type}"
+                    )
                     controller.post_endpoint(endpoint=enable_disable_endpoint)
                     module.exit_json(changed=True)
 
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -24,6 +24,15 @@ options:
       - The name of the rulebook activation.
     type: str
     required: true
+  new_name:
+    description:
+      - Setting this option will change the existing name.
+    type: str
+  copy_from:
+    description:
+      - Name to copy the rulebook activation from.
+      - This will copy an existing rulebook activation and change any parameters supplied.
+      - The new rulebook activation name will be the one provided in the name parameter.
   description:
     description:
       - The description of the rulebook activation.
@@ -56,7 +65,7 @@ options:
     description:
       - Whether the rulebook activation is enabled or not.
     type: bool
-    default: true
+    default: false
   decision_environment_name:
     description:
       - The name of the decision environment associated with the rulebook activation.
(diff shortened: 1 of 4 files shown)


plugins/modules/rulebook_activation.py
```diff
@@ -518,20 +616,69 @@ def main() -> None:
...
-            changed=False,
-            id=activation["id"],
-        )
-
     # Activation Data that will be sent for create/update
     activation_params = create_params(module, controller, is_aap_24=is_aap_24)
-    activation_params["name"] = name
+    activation_params["name"] = new_name if new_name else name
+
+    # Handle copies before anything else
```
alinabuzachis, line 623: I would suggest creating two dedicated functions for these two new features and calling them in the main to keep the main function shorter. It will also be useful to understand if any of these can be moved into module_utils directly.

plugins/modules/rulebook_activation.py
```diff
@@ -49,14 +59,14 @@
...
-    default: "on-failure"
     choices: ["on-failure", "always", "never"]
     type: str
   enabled:
     description:
       - Whether the rulebook activation is enabled or not.
+      - This field will be removed in version 3.0.0.
     type: bool
-    default: true
+    default: false
```
bzwei, line 69: why do we want to change the default to false? In api it defaults to true

plugins/modules/rulebook_activation.py
```diff
@@ -84,6 +94,7 @@
       - A list of IDs for EDA credentials used by the rulebook activation.
       - This parameter is supported in AAP 2.5 and onwards.
         If specified for AAP 2.4, value will be ignored.
+      - Required when state is present.
```
bzwei, line 97: eda_credentials are not required regardless the state

plugins/modules/rulebook_activation.py
```diff
@@ -164,13 +172,62 @@
...
+    project_name: "Example Project"
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
+
+- name: Rename rulebook activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation - copy"
+    new_name: "Renamed Example Rulebook Activation"
+    project_name: "Example Project"
```
bzwei, line 208: project_name and the below are not required for renaming

plugins/modules/rulebook_activation.py
```diff
@@ -164,13 +172,62 @@
...
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
+
+- name: Enable activation
+  ansible.eda.rulebook_activation:
+    name: "Renamed Example Rulebook Activation"
+    state: enabled
+    project_name: "Example Project"
```
bzwei, line 217: project_name and below are not required for enabling

plugins/modules/rulebook_activation.py
```diff
@@ -164,13 +172,62 @@
...
+    rulebook_name: "hello_controller.yml"
+    decision_environment_name: "Example Decision Environment"
+    organization_name: "Default"
+
+- name: Disable activation
+  ansible.eda.rulebook_activation:
+    name: "Renamed Example Rulebook Activation"
+    state: disabled
+    project_name: "Example Project"
```
bzwei, line 226: project_name and the rest are not required for disabling

plugins/modules/rulebook_activation.py
```diff
@@ -518,20 +616,69 @@ def main() -> None:
...
+                # while it's still enabled. We simply honor the operation first before
+                # anything else.
+                if op_type != NO_OP:
+                    activation_id = activation["id"]
+                    enable_disable_endpoint = f"activations/{activation_id}/{op_type}"
+                    controller.post_endpoint(endpoint=enable_disable_endpoint)
+                    module.exit_json(changed=True)
+
+                result = controller.update_if_needed(
+                    activation,
```
bzwei, line 671: if the update is after copying, the id of activation should be updated with the new id, so the update will apply to the copied activation.

plugins/modules/rulebook_activation.py
```diff
@@ -418,29 +477,57 @@ def create_params(
     if not is_aap_24 and module.params.get("log_level"):
         activation_params["log_level"] = module.params["log_level"]
 
-    if not is_aap_24 and module.params.get("swap_single_source") is not None:
-        activation_params["swap_single_source"] = module.params["swap_single_source"]
```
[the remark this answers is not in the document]
kaiokmo: since it is marked for deprecation and we will be removing it in 3.0.0, should we add another check similar to `is_aap_24` or something? in this case, how can this parameter be ignored?

plugins/modules/rulebook_activation.py
```diff
@@ -131,19 +142,15 @@
...
         If specified for AAP 2.4, value will be ignored.
     type: str
-    default: "error"
     choices: ["debug", "info", "error"]
   state:
     description:
       - Desired state of the resource.
     default: "present"
-    choices: ["present", "absent"]
+    choices: ["present", "absent", "enabled", "disabled"]
```
Alex-Izquierdo, line 150: Since we allow the edition of an activation, the behavior of the module changes significantly and we need to explain in detail the behavior. We have to state that "present" is equivalent to "enabled" unless future deprecated "enabled" param is defined (the default is True). Also we have to say that in "present", "enabled" and "disabled" states, the activation will be created if it not exists.

plugins/modules/rulebook_activation.py
```diff
@@ -164,13 +172,62 @@
...
-    enabled: false
-    awx_token_name: "Example Token"
     organization_name: "Default"
+    eda_credentials:
+        - My AAP Credentials
     event_streams:
       - event_stream: "Example Event Stream"
         source_name: "Sample source"
 
+- name: Update rulebook activation fields
```
Alex-Izquierdo, line 182: We need another example to create an activation disabled, for the new interface.

plugins/modules/rulebook_activation.py
Alex-Izquierdo: @bzwei with the current logic, it would be required. 
The point is that with "present", "enabled" and "disabled" you should create the activation if it not exists. So far, any parameter required for the creation (project, org, de, etc...)  are required for "present" state and should be also for the new "enabled" and "disabled"

I see two options, I'm not sure which is the best. 
1. Keep the current logic and require mandatory params for creation for these states
2. Check the existence of required params for creation and raise an error only if the activation does not already exist. 

@alinabuzachis I would like to know your recommendation. How is handle this situation in other collections?

plugins/modules/rulebook_activation.py
```diff
@@ -455,9 +542,16 @@ def main() -> None:
...
-        state=dict(choices=["present", "absent"], default="present"),
+        swap_single_source=dict(
+            type="bool",
+            default=True,
+            removed_in_version="3.0.0",
+            removed_from_collection="ansible.eda",
+        ),
+        log_level=dict(type="str", choices=["debug", "info", "error"]),
+        state=dict(
+            choices=["present", "absent", "enabled", "disabled"], default="present"
```
Alex-Izquierdo, line 553: As I elaborated in a previous comment, the new default should be "enabled" because the default of the api param "is_enabled" is True. I mean, by default the activation is created enabled.

plugins/modules/rulebook_activation.py
```diff
@@ -24,6 +24,16 @@
...
     required: true
+  new_name:
+    description:
+      - Setting this option will change the existing name.
+    type: str
+  copy_from:
+    description:
+      - Name to copy the rulebook activation from.
+      - This will copy an existing rulebook activation and change any parameters supplied.
+      - The new rulebook activation name will be the one provided in the name parameter.
```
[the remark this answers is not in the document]
Alex-Izquierdo: Regardless of my suggestion, I noticed that `change any parameters supplied` statement is not true. 
You only pass the new name, call the copy endpoint and finalizing the task.
I understand that the behavior we want is to edit the copied activation with any other parameter supplied. 
If want this behavior, the module needs more logic to update the copied activation. 
Otherwise this documentation is wrong.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -172,14 +201,296 @@
...
+          - _result.changed
+
+    - name: Get information about rulebook activation
+      ansible.eda.rulebook_activation_info:
+      register: _result_activation_info
+
+    - name: Assert that copied rulebook activations are equal except name
+      ansible.builtin.assert:
+        that:
+          - _result_activation_info.activations[0].description  ==
```
Alex-Izquierdo, line 226: The test assumes that there are not other activation created. This is a bad practice. It should fetch specifically the expected activations and not depend on the position in the list.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -172,14 +201,296 @@
         that:
           - _result.changed
 
+    - name: Copy rulebook activation
+      ansible.eda.rulebook_activation:
+        name: "{{ activation_name }}-copy"
+        copy_from: "{{ activation_name }}"
+        project_name: "{{ project_name }}"
```
Alex-Izquierdo, line 208: As it is now no other param than the new_name is used by the copy operation, so they should not be needed. Unless we want to honor what it is described in the documentation and update the copied activation with the other params provided.  In this case, it should have a different project_name, to be able to check that the copied activation already has the new project and not the original. 

As it is now, checking if it is has the same project is unnecessary, you would be just testing the backend when copying, which is out of the scope of this test.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -172,14 +201,296 @@
...
+
+    - name: Get information about rulebook activation
+      ansible.eda.rulebook_activation_info:
+      register: _result_activation_info
+
+    - name: "Extract activation with name {{ activation_name }}-suffix"
+      set_fact:
+        filtered_activation: "{{ _result_activation_info.activations | selectattr('name', 'equalto', activation_name + '-suffix') | list | first }}"
+
+    - name: Assert that rulebook activation is updated
```
Alex-Izquierdo, line 336: Why not simplify these tests and update all the fields in only one task?

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -172,14 +201,296 @@
...
+      register: _result
+      ignore_errors: true
+
+    - name: Check rulebook activation creation
+      assert:
+        that:
+          - _result.failed
+          - "'The rulebook requires a RH AAP credential.' in _result.msg"
+
+    - name: Change name to existing activation with same name
```
Alex-Izquierdo, line 408: Which module feature are you testing here? I think this test is not necessary. You are just testing the backend.

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -172,14 +201,296 @@
...
+      register: _result
+      ignore_errors: true
+
+    - name: Check rulebook activation creation
+      assert:
+        that:
+          - _result.failed
+          - "'activation with this name already exists' in _result.msg"
+
+    - name: Update activation in enabled state
```
Alex-Izquierdo, line 427: Which module feature are you testing here? I think this test is not necessary. You are just testing the backend.

Alex-Izquierdo (commented):
@kaiokmo What is the plan for the "wait" logic that would be necessary here when enabling/disabling the activation? (and also be necessary for the future restart of Brandon

plugins/modules/rulebook_activation.py
Alex-Izquierdo: I think all these parameters should not be mandatory unless we need it to create the activation. 
For example, I use the podman module to run a container:
```
- hosts: localhost
  connection: local
  tasks:
  - name: Run container
    containers.podman.podman_container:
      name: container
      image: localhost/aap-eda
      command: sleep 1000
      state: started
```

And I run the task again but only with the name, the module does not fail, because the container is in the expected state:
```
- hosts: localhost
  connection: local
  tasks:
  - name: Run container
    containers.podman.podman_container:
      name: container
      # image: localhost/aap-eda
      # command: sleep 1000
      state: started
```

If I delete the pod I try to run the task without image, is when the module handles the missing param:
```
TASK [Run container] **********************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Cannot start container when image is not specified!"}
``` 

This is how I think we should behave.

plugins/modules/rulebook_activation.py
```diff
@@ -518,20 +616,69 @@ def main() -> None:
...
+                # we are not directly managing the state of an activation.
+                #
+                # This doesn't handle cases where the user tries to update an activation
+                # while it's still enabled. We simply honor the operation first before
+                # anything else.
+                if op_type != NO_OP:
+                    activation_id = activation["id"]
+                    enable_disable_endpoint = f"activations/{activation_id}/{op_type}"
+                    controller.post_endpoint(endpoint=enable_disable_endpoint)
+                    module.exit_json(changed=True)
```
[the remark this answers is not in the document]
Alex-Izquierdo: Are operations in the sense that are operations in the backend (they have a concrete endpoint to trigger an async task)
We also are going to implement "wait" logic, to wait for the expected state within a timeout. 
But it is a state in the sense that the module should be ok if the state is "enabled" and the rulebook already exists as "enabled"

plugins/modules/rulebook_activation.py
alinabuzachis: @Alex-Izquierdo I would go for option #1 which is also similar to the approach you are showing in the example, if I understand correctly.

Alex-Izquierdo: @alinabuzachis It is the opposite, what I observe in the example with podman module is the second option. We take care of the state of the resource and we only require the parameter when it is really needed. For example, if the activation already exists, the DE would not be required.

kaiokmo:
Abandoning in favor of #400

PR #398, opened 2025-02-24
fix: add support for Kafka GSSAPI

mkanoor:
The aiokafka package already supports GSSAPI as long as it has the correct packages installed along with kerberos.

This fix just passes in the parameters from EDA to aiokafka

The DE should have the following Kerberos packages on CentOS:
 
   * krb5-libs
   * krb5-devel

Along with the python gssapi package.

https://issues.redhat.com/browse/AAPRFE-1662

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,10 @@ requires =
 ignore_basepython_conflict = false
 
 [testenv]
+# Currently ansible has issues loading Cython compiled modules
+# from under the ansible_collection/ansible/eda directory.
+# https://github.com/ansible/ansible/pull/84758
+env_dir=/tmp/event_driven_ansible/.tox/{env_name}
 description =
   Run unit tests
   py{39,310,311,312,313}: with {basepython}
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo:
> The DE should have
> 
> * krb5-workstation
> * krb5-libs
> * krb5-devel

It would be good to install it in ansible-rulebook image. WDYT?

mkanoor:
https://github.com/ansible/ansible/pull/84758

tox.ini
```diff
@@ -23,6 +23,10 @@ requires =
 ignore_basepython_conflict = false
 
 [testenv]
+# Currently ansible has issues loading Cython compiled modules
+# from under the ansible_collection/ansible/eda directory.
+# https://github.com/ansible/ansible/pull/84758
+env_dir=/tmp/hack/.tox/{env_name}
```
Alex-Izquierdo, line 29: Can we remove the "hack" term and just use some reference to the project?

Alex-Izquierdo: happy to see the solution is so easy.

Alex-Izquierdo (approved):

Dostonbek1 (approved):

hsong-rh (approved):

PR #399, opened 2025-02-24
feat: [AAP-37428] add support for copy credential

Dostonbek1:
This PR adds support for EDA Credentials' new copy endpoint - `/eda-credentials/<id>/copy`.

### DIFF
--- a/plugins/modules/credential.py
+++ b/plugins/modules/credential.py
@@ -36,6 +36,7 @@ options:
       - If set, copies the specified credential.
       - The new credential will be created with the name given in the C(name) parameter.
     type: str
+    version_added: '3.0.0'
   inputs:
     description:
       - Credential inputs where the keys are var names used in templating.
@@ -192,7 +193,6 @@ def main() -> None:
 
     name = module.params.get("name")
     new_name = module.params.get("new_name")
-    copy_from = module.params.get("copy_from")
     state = module.params.get("state")
 
     # Attempt to look up credential based on the provided name
@@ -221,19 +221,22 @@ def main() -> None:
         else (controller.get_item_name(credential) if credential else name)
     )
 
+    # we attempt to copy only when copy_from is passed
     if copy_from:
-        if not credential:
-            module.fail_json(msg=f"Credential with name {copy_from} was not found.")
-        else:
-            try:
-                credential_id = credential["id"]
-                copy_endpoint = f"eda-credentials/{credential_id}/copy"
-                params = {"name": name}
-
-                controller.post_endpoint(endpoint=copy_endpoint, data=params)
-                module.exit_json(changed=True)
-            except EDAError as e:
-                module.fail_json(msg=f"Failed to copy credential: {e}")
+        try:
+            result = controller.copy_if_needed(
+                name,
+                copy_from,
+                endpoint=f"eda-credentials/{credential['id']}/copy",
+                item_type="credential",
+            )
+            module.exit_json(changed=True)
+        except KeyError as e:
+            module.fail_json(
+                msg=f"Unable to access {e} of the credential to copy from."
+            )
+        except EDAError as e:
+            module.fail_json(msg=f"Failed to copy credential: {e}")
 
     # If the state was present and we can let the module build or update the
     # existing credential, this will return on its own
--- a/plugins/modules/credential.py
+++ b/plugins/modules/credential.py
@@ -30,6 +30,12 @@ options:
     description:
       - Setting this option will change the existing name (lookup via name).
     type: str
+  copy_from:
+    description:
+      - Name of the existing credential to copy.
+      - If set, copies the specified credential.
+      - The new credential will be created with the name given in the C(name) parameter.
+    type: str
   inputs:
     description:
       - Credential inputs where the keys are var names used in templating.
@@ -75,6 +81,11 @@ EXAMPLES = r"""
   ansible.eda.credential:
     name: "Example Credential"
     state: absent
+
+- name: Copy an existing EDA Credential
+  ansible.eda.credential:
+    name: "New Copied Credential"
+    copy_from: "Existing Credential"
 """
 
 
@@ -136,6 +147,7 @@ def main() -> None:
     argument_spec = dict(
         name=dict(type="str", required=True),
         new_name=dict(type="str"),
+        copy_from=dict(type="str"),
         description=dict(type="str"),
         inputs=dict(type="dict"),
         credential_type_name=dict(type="str"),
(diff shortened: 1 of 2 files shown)


plugins/modules/credential.py
```diff
@@ -201,6 +221,20 @@ def main() -> None:
         else (controller.get_item_name(credential) if credential else name)
     )
 
+    if copy_from:
+        if not credential:
```
Alex-Izquierdo, line 225: WDYT if we encapsulate the copy_from logic in a function?

Dostonbek1: I think that's a good idea. I tried to follow the existing pattern, let me know what you think.

plugins/modules/credential.py
```diff
@@ -201,6 +221,19 @@ def main() -> None:
...
+    try:
+        result = controller.copy_if_needed(
+            name,
+            copy_from,
+            endpoint=f"eda-credentials/{credential['id']}/copy",
+            item_type="credential",
+        )
+        module.exit_json(changed=True)
+    except KeyError:
+        module.fail_json(msg="Unable to access ID of the credential to copy from: {e}")
```
kaiokmo, line 233: I think this could break the module if an unexisting credential is passed in the `copy_from` field, as `e` would never exist. So I guess we should either bind the exception to `e` or remove it from the fail message.

plugins/modules/credential.py
```diff
@@ -175,11 +192,14 @@ def main() -> None:
 
     name = module.params.get("name")
     new_name = module.params.get("new_name")
+    copy_from = module.params.get("copy_from")
```
kaiokmo, line 195: Is this duplicate intended? What's the difference of this to the value of `copy_from` we already got in the line `161`?

kaiokmo (changes requested):

plugins/modules/credential.py
Dostonbek1: I swear I thought I added e in the exception. But updated it now, the output looks like the following
```

TASK [Copy an existing EDA Credential] ****************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to access 'id' of the credential to copy from"}
```

plugins/modules/credential.py
Dostonbek1: good catch. Updated.

plugins/modules/credential.py
```diff
@@ -201,6 +220,19 @@ def main() -> None:
         else (controller.get_item_name(credential) if credential else name)
     )
 
+    try:
```
alinabuzachis, line 223: I think this should not fail when the user does not want to copy anything so copy_from is not set. If I understand correctly, it will fail right now if I do not set copy_from.

Dostonbek1: Updated to check for copy_from param, before we get into this try block.

plugins/module_utils/controller.py
```diff
@@ -383,3 +383,31 @@ def delete_if_needed(
...
+        self,
+        name: str,
+        copy_from: str,
+        endpoint: str,
+        item_type: str = "unknown",
+    ) -> dict[str, bool]:
+        response = None
+        if not copy_from:
+            msg = f"Unable to copy {item_type}, missing copy_from parameter"
+            raise EDAError(msg)
```
alinabuzachis, line 397: I think this can be removed because copy_if_needed() will be called only when copy_from is defined.

plugins/modules/credential.py
```diff
@@ -201,6 +220,23 @@ def main() -> None:
...
     )
 
+    # we attempt to copy only when copy_from is passed
+    if copy_from:
+        try:
+            result = controller.copy_if_needed(
+                name,
+                copy_from,
+                endpoint=f"eda-credentials/{credential['id']}/copy",
+                item_type="credential",
```
alinabuzachis, line 230: This is only a suggestion. Since your passing item_type to copy_if_needed(), why don't you build the endpoint inside the function based on the item_type?

For example: if item_type == "credential" -> endpoint = f"eda-credentials/{id}/copy" elif item_type == "activation" -> endpoint = f"activations/{activation['id']}/copy"

You will have copy_if_needed(name, copy_from, id, item_type).

Dostonbek1: I think moving the endpoint into the copy_if_needed function would be fine for now, but thinking for future scenarios it might get difficult to scale it, for example, if we add copy feature for other resources then we will end up with a big block of "if elif" code. Thoughts?

kaiokmo: I agree with Doston. In this case, we would end up having a big if elif block.

What we can do is to pass the endpoint as plain `endpoint="eda-credentials"` to `controller.copy_if_needed`, and, inside the function, retrieve the resource id with something like `id = controller.get_exactly_one` and finally format the endpoint like `endpoint = '{0}/{1}/{2}'.format(endpoint, id, "copy")` before sending the request `response = self.post_endpoint(endpoint, data={"name": name})`.

That way we keep the `item_type` behavior for informational purposes and guarantee it will work for other resources.

WDYT?

alinabuzachis: get_exactly_one()  is already called here https://github.com/ansible/event-driven-ansible/pull/399/files#diff-e8e6c3361f7f035f239fecccb755edfa684f11c90374ecf6abac2cb11b61d408R199, so it will be called two times then.
I am fine with the implementation as it is and we can think about improvement later. I don't want to block the merging of this PR.

alinabuzachis (approved):

plugins/modules/credential.py
```diff
@@ -30,6 +30,12 @@
     description:
       - Setting this option will change the existing name (lookup via name).
     type: str
+  copy_from:
+    description:
+      - Name of the existing credential to copy.
+      - If set, copies the specified credential.
+      - The new credential will be created with the name given in the C(name) parameter.
```
alinabuzachis, line 37: Can you please add  version_added: 3.0.0?

plugins/module_utils/controller.py
```diff
@@ -387,13 +387,14 @@ def delete_if_needed(
...
-        copy_from: str,
+        copy_from_name: str,
+        copy_from_id: Union[int, Any],
         endpoint: str,
         item_type: str = "unknown",
     ) -> dict[str, bool]:
         response = None
-        if not copy_from:
-            msg = f"Unable to copy {item_type}, missing copy_from parameter"
+        if not copy_from_id:
```
Alex-Izquierdo, line 396: Why duplicate the params? If you are going to refactor this, why not just use the id?

plugins/modules/credential.py
```diff
@@ -145,13 +158,18 @@ def main() -> None:
...
     argument_spec.update(AUTH_ARGSPEC)
 
-    required_if = [
-        (
-            "state",
-            "present",
-            ("name", "credential_type_name", "inputs", "organization_name"),
-        )
-    ]
+    module = AnsibleModule(argument_spec=argument_spec)
```
kaiokmo, line 161: Can you please also pass `supports_check_mode=True` to the constructor?

kaiokmo (changes requested):

plugins/modules/credential.py
Dostonbek1: added.

alinabuzachis (approved):

kaiokmo (approved):

plugins/modules/credential.py
Alex-Izquierdo: We are going to release 2.6, not 3.0. @alinabuzachis I think it is a mistake. No major release planned, this is not a breaking change.

PR #402, opened 2025-02-27
Fix YAML syntax issues in DOCUMENTATION

sivel:
A few of the `event_source` plugins had invalid DOCUMENTATION due to YAML syntax errors, this PR addresses them by using folded or block scalars.

Generally, if there is a `:` in the string, with whitespace to the right, YAML will split it as a key/value pair. Using folded or block scalars will resolve the issue as the whole field will be treated as a string.

### DIFF
--- a/extensions/eda/plugins/event_source/aws_cloudtrail.py
+++ b/extensions/eda/plugins/event_source/aws_cloudtrail.py
@@ -11,7 +11,8 @@ DOCUMENTATION = r"""
 short_description: Receive events from an AWS CloudTrail
 description:
   - An ansible-rulebook event source module for getting events from an AWS CloudTrail.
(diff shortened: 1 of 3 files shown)


ttuffin (approved):

sivel:
fwiw, these were found using the following commands, that kind of abose the `-M` argument for `ansible-doc`:

```
for F in `ls event_filter/*.py`; do ansible-doc -M event_filter -j "$(basename $F)"; done
for F in `ls event_source/*.py`; do ansible-doc -M event_source -j "$(basename $F)"; done
```

alinabuzachis (approved):

Dostonbek1 (approved):

zkayyali812 (approved):

PR #409, opened 2025-03-12
feat [AAP-40577]: add support for edit activations

kaiokmo:
https://issues.redhat.com/browse/AAP-40577

This also covers enabling and disabling rulebook activations, as well
as updating other fields like name, log_level, credentials, and so on.

### DIFF
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -28,14 +28,14 @@ options:
       - Renames an existing rulebook activation.
       - If set, the rulebook activation will be updated with the new name.
     type: str
-    version_added: 2.6.0
+    version_added: 2.7.0
   copy_from:
     description:
       - Name of the existing rulebook activation to copy.
       - If set, copies the specified rulebook activation.
       - The new rulebook activation will be created with the name given in the C(name) parameter.
     type: str
-    version_added: 2.6.0
+    version_added: 2.7.0
   description:
     description:
       - The description of the rulebook activation.
@@ -74,6 +74,9 @@ options:
   enabled:
     description:
       - Whether the rulebook activation is enabled or not.
+      - This field will be removed in version 3.0.0.
+      - The logic of controlling the state of an activation is going to be controlled
+        by the C(state) parameter itself.
     type: bool
     default: true
   decision_environment_name:
@@ -120,7 +123,6 @@ options:
       - This field will be removed in version 3.0.0.
     type: bool
     default: true
-    removed_in: 3.0.0
   event_streams:
     description:
       - A list of event stream names that this rulebook activation listens to.
@@ -156,14 +158,19 @@ options:
   state:
     description:
       - Desired state of the resource.
+      - The state of an activation itself is controlled by this parameter, whether I(enabled) or
+        I(disabled).
+      - When an activation is created, it is I(enabled) by default. Thus, I(present) is equivalent to
+        I(enabled).
+      - Whether I(present), I(enabled), or I(disabled) is specified, an activation will be created if it
+        doesn't exist already.
+      - To create a disabled activation, specify C(state) as I(disabled).
+      - Essentially, this parameter deprecates the usage of C(enabled).
     default: "present"
-    choices: ["present", "absent"]
+    choices: ["present", "absent", "disabled", "enabled"]
     type: str
 extends_documentation_fragment:
   - ansible.eda.eda_controller.auths
-notes:
-  - Rulebook Activation API does not support PATCH method, due to this reason the module will
-    not perform any modification when an existing rulebook activation is found.
 """
 
 EXAMPLES = r"""
@@ -174,8 +181,7 @@ EXAMPLES = r"""
     project_name: "Example Project"
     rulebook_name: "hello_controller.yml"
     decision_environment_name: "Example Decision Environment"
-    enabled: false
-    awx_token_name: "Example Token"
+    state: disabled
 
 - name: Create a rulebook activation with event_streams option
   ansible.eda.rulebook_activation:
@@ -184,8 +190,7 @@ EXAMPLES = r"""
     project_name: "Example Project"
     rulebook_name: "hello_controller.yml"
     decision_environment_name: "Example Decision Environment"
-    enabled: false
-    awx_token_name: "Example Token"
+    state: disabled
     organization_name: "Default"
     event_streams:
       - event_stream: "Example Event Stream"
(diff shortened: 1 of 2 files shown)


plugins/modules/rulebook_activation.py
```diff
@@ -24,6 +24,12 @@
       - The name of the rulebook activation.
     type: str
     required: true
+  new_name:
+    description:
+      - Renames an existing rulebook activation.
+      - If set, the rulebook activation will be updated with the new name.
+    type: str
+    version_added: 2.6.0
```
Alex-Izquierdo, line 32: We already released 2.6.0, this should be 2.7.0, right?

Alex-Izquierdo (commented):
This needs to handle the mandatory params differently. All the paramaters required for a creation should be required only if we are going to create it. Look at this discussion as reference: https://github.com/ansible/event-driven-ansible/pull/394#discussion_r1967861910

A task with only "state=present and name" should be valid if the activation already exists. Otherwise it should fail because we need more parameters to create the activation. 
With the edition supported, any parameter passed is going to be passed for a patch if the activation exists. I think we miss some tests to validate all the possible scenarios.

plugins/modules/rulebook_activation.py
```diff
@@ -555,22 +601,39 @@ def main() -> None:
...
+    if state in ("enabled", "disabled") and activation:
+        activation_params["is_enabled"] = True if state == "enabled" else False
+
+        # Handle disable separately, as enable is treated by the backend.
+        try:
+            if activation["is_enabled"] and not activation_params["is_enabled"]:
+                controller.post_endpoint(
+                    endpoint=f"activations/{activation['id']}/disable"
+                )
+                module.exit_json(changed=True)
```
bzwei, line 628: After the activation is disabled, should you always exit the module or continue if there are other attributes to be updated in `activation_params`?

plugins/modules/rulebook_activation.py
```diff
@@ -143,14 +154,18 @@
   state:
     description:
       - Desired state of the resource.
+      - The state of an activation itself is controlled by this parameter, whether enabled or
+        disabled.
+      - When an activation is created, it is enabled by default. Thus, present is equivalent to
```
bzwei, line 159: Since param `enabled` is deprecated, how can a user create an activation in disabled mode?
Your code does not force activation_param["is_enabled"] = True when user does not provide enable param but set state to "present".

plugins/modules/rulebook_activation.py
kaiokmo: You're, I will adjust it.

kaiokmo: You're right. I'll adjust it.

plugins/modules/rulebook_activation.py
kaiokmo: Yes, it goes with the backend defaults. But it's not a bad idea to enforce it to true it when state is present. I will document both scenarios.

plugins/modules/rulebook_activation.py
kaiokmo: @bzwei This is actually a design decision. Whether we:
  - treat them as separate things and perform enable/disable, and edit, in different tasks
  - allow an activation to be enabled/disabled/edited in one shot, which honors the state management aspect of the module that we want to have

cc: @Alex-Izquierdo @alinabuzachis I believe we talked about this, but given we are going for stateful management of activations, we should stick with option no. 2, right?

Alex-Izquierdo: We don't want to be operational anymore. The task declares a state and the module should ensure the activation is the declared state unless the declaration is invalid.  The point here is that the activation requires to be disabled to be edited. So once the state is disabled is should be updated according to what it is declared in the task. 

If you try to edit an activation that it is in enabled state the backend would return and error and the module should fail with that error.

kaiokmo: @Alex-Izquierdo So if one tries to disable and edit the activation at the same time, it's a valid state, right? The module will have to disable it and then send the update.

What if the activation takes a bit longer to be disabled and the module tries to perform the update but when it does, the activation is still not disabled? We would need to wait for the disabling to go through, right? Otherwise, the backend would return an error

plugins/modules/rulebook_activation.py
Alex-Izquierdo: @kaiokmo I think for now it must be a two-operations thing because we internally discussed to postpone the wait for logic for later iterations. And this behavior would require the wait for logic since we need to disable the activation and wait to be disabled before editing it.

plugins/modules/rulebook_activation.py
bzwei: To allow this combination, we need the wait_for implementation, either a default timeout or user configurable. If wait_for is a separate feature, this PR should not support disable and editing combination.

Alex-Izquierdo: Yeah, is the same that I said, just I'm stupid and did it in the wrong thread. :(
https://github.com/ansible/event-driven-ansible/pull/409#discussion_r2010392991

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -360,44 +384,106 @@
...
-        state: absent
+        state: disabled
+
+    - name: Check rulebook activation not updated
+      assert:
+        that:
+          - not _result.changed
+
+    - name: Update rulebook activation name and fields
+      ansible.eda.rulebook_activation:
```
zkayyali812, line 435: @kaiokmo Can we add or update a test where we modify the event streams field?

kaiokmo: done

zkayyali812 (commented):
Very minor comment, but looks great!!

plugins/modules/rulebook_activation.py
```diff
@@ -190,6 +204,26 @@
...
+- name: Enable a rulebook activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation"
+    state: enabled
+    organization_name: "Default"
+
+- name: Disable a rulebook activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation"
+    new_name: "Example Rulebook Activation New Name"
```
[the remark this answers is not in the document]
kaiokmo: I added new_name here to illustrate that one could update the name of the activation together with other parameters. I will split it into two distinct examples.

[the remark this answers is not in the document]
kaiokmo: done!

zkayyali812 (approved):

PR #410, opened 2025-03-13
fix: add restart activation AAP-37681

B-Whitt:
Rulebook activation [module](https://console.redhat.com/ansible/automation-hub/repo/published/ansible/eda/content/module/rulebook_activation/) lacks support for restarting the activation. As any existing operation it should be supported. 
Jira: https://issues.redhat.com/browse/AAP-37681

### DIFF
--- a/plugins/module_utils/controller.py
+++ b/plugins/module_utils/controller.py
@@ -411,3 +411,27 @@ class Controller:
             raise EDAError(response.json["__all__"])
         msg = f"Unable to copy from {item_type} {copy_from}: {response.status}"
         raise EDAError(msg)
+
+    def restart_if_needed(
+        self,
+        existing_item: dict[str, Any],
+        endpoint: str,
+    ) -> dict[str, bool]:
+        if not endpoint:
+            msg = "Unable to restart activation, missing endpoint"
+            raise EDAError(msg)
+        if not existing_item:
+            msg = "Unable to restart activation, missing activation parameter"
+            raise EDAError(msg)
+        if self.module.check_mode:
+            return {"changed": True}
+
+        item_id = existing_item["id"]
+        response = self.post_endpoint(endpoint, data={"id": item_id})
+        if response.status in [200, 201, 204]:
+            self.result["changed"] = True
+            return self.result
+        if response.json and "__all__" in response.json:
+            raise EDAError(response.json["__all__"])
+        msg = f"Error restarting activation with ID: {existing_item['id']}"
+        raise EDAError(msg)
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -8,7 +8,6 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-
 DOCUMENTATION = r"""
 ---
 module: rulebook_activation
@@ -53,6 +52,13 @@ options:
     description:
       - The extra variables for the rulebook activation.
     type: str
+  restart:
+    description:
+      - Performs a restart of the activation. This is a non idempotent operation.
+      - If enabled the rest of parameters will be ignored.
+    type: bool
+    default: false
+    version_added: 2.7.0
   restart_policy:
     description:
       - The restart policy for the rulebook activation.
@@ -188,6 +194,11 @@ EXAMPLES = r"""
   ansible.eda.rulebook_activation:
     name: "Example Rulebook Activation"
     state: absent
+
+- name: Restart activation
+  ansible.eda.rulebook_activation:
+    name: "Example Rulebook Activation - Restart"
+    restart: true
 """
 
 
@@ -198,7 +209,6 @@ id:
   type: int
   sample: 37
 """
-
 
 import traceback
 from typing import Any, Dict, List
@@ -445,6 +455,7 @@ def main() -> None:
         project_name=dict(type="str", aliases=["project"]),
         rulebook_name=dict(type="str", aliases=["rulebook"]),
         extra_vars=dict(type="str"),
+        restart=dict(type="bool", default=False),
         restart_policy=dict(
             type="str",
             default="on-failure",
@@ -478,9 +489,10 @@ def main() -> None:
 
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
     copy_from = module.params.get("copy_from", None)
+    restart = module.params.get("restart", None)
 
     required_if = []
-    if not copy_from:
+    if not (copy_from or restart):
         required_if = [
             (
                 "state",
@@ -538,6 +550,16 @@ def main() -> None:
             module.exit_json(**result)
         except EDAError as e:
             module.fail_json(msg=f"Failed to delete rulebook activation: {e}")
+
+    if restart:
+        try:
+            result = controller.restart_if_needed(
+                activation,
+                endpoint=f"activations/{activation['id']}/restart",
+            )
+            module.exit_json(**result)
+        except EDAError as e:
+            module.fail_json(msg=(f"Failed to restart rulebook activation: {e}."))
 
     if copy_from:
         try:
--- a/tests/integration/targets/activation/tasks/activation_lifecycle.yml
+++ b/tests/integration/targets/activation/tasks/activation_lifecycle.yml
@@ -1,0 +1,138 @@
+---
+- name: Rulebook activation integration tests for lifecycle operations
+  module_defaults:
+    group/ansible.eda.eda:
+      aap_hostname: "{{ aap_hostname }}"
+      aap_username: "{{ aap_username }}"
+      aap_password: "{{ aap_password }}"
+      aap_verify_ssl: "{{ aap_verify_ssl }}"
+  block:
+    - name: Generate a random_string for the test
+      set_fact:
+        random_string: "{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
+      when: random_string is not defined
+
+    - name: Generate a ID for the test
+      set_fact:
+        test_id: "{{ random_string | to_uuid }}"
+      when: test_id is not defined
+
+    - name: Define variables for credential and project
+      set_fact:
+        organization_name: "Default"
+        decision_env_name: "Test_Decision_Env_{{ test_id }}"
+        activation_name: "Test_Activation_{{ test_id }}"
+        project_name: "Test_Project_{{ test_id }}"
+        de_image_url: "quay.io/ansible/ansible-rulebook:main"
+        scm_url: https://github.com/ansible/eda-sample-project.git
+        rulebook_name: "range_long_running.yml"
+
+    - name: Create a new project
+      debugger: on_failed
+      ansible.eda.project:
+        name: "{{ project_name }}"
+        description: "Test Project Description"
+        url: "{{ scm_url }}"
+        organization_name: "{{ organization_name}}"
+        state: present
+      register: project_creation
+
+    - name: Wait for the project to be imported
+      register: project_status
+      until: project_status.projects[0].import_state == "completed"
+      retries: 30
+      delay: 2
+      ansible.eda.project_info:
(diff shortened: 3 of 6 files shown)


plugins/modules/rulebook_activation.py
```diff
@@ -53,6 +73,27 @@
...
+      - Whether the system should wait after an activation restart.
+    type: bool
+    default: false
+  restart_wait_time:
+    description:
+      - Represents the total number of seconds that the system waits,
+        after an activation restart, before it returns.
+    type: int
+    default: 200
+  restart_interval:
```
bzwei, line 92: The name `restart_interval` is misleading. It sounds like the interval between multiple restarts. It is actually the check status interval. To me user does not need to set the interval.

plugins/modules/rulebook_activation.py
```diff
@@ -53,6 +73,27 @@
...
+    description:
+      - Whether or not the activation should be restarted.
+    type: bool
+    default: false
+  restart_wait:
+    description:
+      - Whether the system should wait after an activation restart.
+    type: bool
+    default: false
+  restart_wait_time:
```
bzwei, line 86: is `restart_timeout` more understandable?

plugins/modules/rulebook_activation.py
```diff
@@ -539,6 +597,27 @@ def main() -> None:
...
+            time.sleep(10)
+            while time.time() < end_time:
+                activation = update_activation(activation, module, controller)
+                if (
+                    activation["status"] == "running"
+                    or activation["status"] == "completed"
+                ):
+                    break
+                time.sleep(restart_interval)
+            result["msg"] = f"Current Activation details: {activation}"
```
bzwei, line 615: What is the purpose of waiting after restart? Isn't that to make sure the activation is restarted and back to running state? So shouldn't we exit the module with error if we cannot detect this status within the timeout window? But there might be cases that the timeout is set too short and the activation still reaches the desired status after timeout.

plugins/modules/rulebook_activation.py
```diff
@@ -437,6 +465,31 @@ def create_params(
     return activation_params
 
 
+def restart_activation(
```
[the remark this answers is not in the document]
B-Whitt: Sounds good. I had it that way and then pulled it out hoping to simplify the code a bit. Ill move it back inside. Thanks

plugins/modules/rulebook_activation.py
B-Whitt: updated to 'restart_timeout'

plugins/modules/rulebook_activation.py
```diff
@@ -437,6 +465,31 @@ def create_params(
...
+            endpoint=f"activations/{item['id']}/restart",
+        )
+    except EDAError as e:
+        module.fail_json(
+            msg=f"Failed to restart rulebook activation: {e} Current Activation restart_count: {item['restart_count']}"
+        )
+    return result
+
+
+def update_activation(
```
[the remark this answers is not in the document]
B-Whitt: updated to 'retrieve_activation'

plugins/modules/rulebook_activation.py
```diff
@@ -53,6 +73,27 @@
     description:
       - The extra variables for the rulebook activation.
     type: str
+  restart:
+    description:
+      - Whether or not the activation should be restarted.
+    type: bool
+    default: false
+  restart_wait:
```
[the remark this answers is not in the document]
B-Whitt: we included the 'restart_wait' to allow the user to wait and use the default 'restart_timeout'

plugins/modules/rulebook_activation.py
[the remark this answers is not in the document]
B-Whitt: moved back inside of main() function

plugins/modules/rulebook_activation.py
Alex-Izquierdo: Agree @bzwei The timeout is set by the playbook author and if it does not reach the expected state in the time window it should fail. This a convenience feature to avoid to the user to write their own loop and it is aligned with other modules with similar features like projects in AAP. Projects are cancelled after the timeout but they are one shot actions, so it makes sense, in this case I think it is better just to raise the error. 

For job templates there is [a specific module](https://docs.ansible.com/ansible/latest/collections/awx/awx/job_wait_module.html#ansible-collections-awx-awx-job-wait-module) to do this. We can consider to use a new module in later iterations.

plugins/modules/rulebook_activation.py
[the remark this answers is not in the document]
Alex-Izquierdo: This interface would not align with the usual pattern and it complicates unnecessarily the interface. 
There is no need for "restart_wait"
We would have "wait", "timeout" and "interval" and those are commons for the activation, regardless of restart, enabled (running) or disabled (stopped) states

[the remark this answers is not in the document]
Alex-Izquierdo: Wait is a boolean which trigger the "wait" functionality. And if wait is set to true, then we need the timeout and interval. 

I think to simplify the implementation, we can just run the restart op and implement the wait logic in a different PR. 
WDTY ? @B-Whitt @bzwei @kaiokmo @zkayyali812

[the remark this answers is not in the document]
bzwei: Agree. We can begin with a restart only PR.

plugins/modules/rulebook_activation.py
bzwei: It is tricky what to do after timeout. We agree it should fail the module. Because there is no remedy after timeout, the only choice seems just do nothing 😄

plugins/modules/rulebook_activation.py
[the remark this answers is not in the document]
B-Whitt: wait logic will be removed and added to a new PR

plugins/modules/rulebook_activation.py
```diff
@@ -53,6 +52,11 @@
     description:
       - The extra variables for the rulebook activation.
     type: str
+  restart:
+    description:
+      - Whether or not the activation should be restarted.
+    type: bool
+    default: false
```
alinabuzachis, line 59: Please add `version_added: 2.7.0`, if the next release is 2.7.0.

plugins/module_utils/controller.py
```diff
@@ -8,13 +8,16 @@
...
 import json
+import logging
 from typing import Any, List, NoReturn, Optional
 
 from ansible.module_utils.basic import AnsibleModule
 
 from .client import Client, Response
 from .errors import EDAError
 
+LOGGER = logging.getLogger(__name__)
```
alinabuzachis, line 19: I think this can be removed, right?

tests/integration/targets/activation/tasks/activation_lifecycle.yml
```diff
@@ -0,0 +1,125 @@
...
+      ansible.eda.rulebook_activation_info:
+        name: "{{ activation_name }}"
+
+    - name: Check if activation was restarted
+      ansible.builtin.assert:
+        that:
+          - activation_status.activations[0].restart_count == 1
+          - activation_status.activations[0].status in ["running", "starting"]
+
+    - name: Restart activation without activation name
```
alinabuzachis, line 93: Would you add tests for check_mode and idempotency?

alinabuzachis (approved):

plugins/module_utils/controller.py
```diff
@@ -8,13 +8,16 @@
 
 
 import json
+import logging
```
alinabuzachis, line 11: Same?

B-Whitt: removed

plugins/modules/rulebook_activation.py
B-Whitt: added

plugins/modules/rulebook_activation.py
```diff
@@ -539,6 +549,20 @@ def main() -> None:
         except EDAError as e:
             module.fail_json(msg=f"Failed to delete rulebook activation: {e}")
 
+    if restart:
+        try:
+            result = controller.restart_if_needed(
+                activation,
+                endpoint=f"activations/{activation['id']}/restart",
+            )
```
bzwei, line 557: Shouldn't we exit the module here? We should document that when restart is true, all other arguments except the name are ignored.

B-Whitt: yes. I caught that earlier. But, thanks for the reminder.

tests/integration/targets/activation/tasks/activation_lifecycle.yml
B-Whitt: both added

Alex-Izquierdo: @alinabuzachis I might be missing something, but restart is a request that triggers an operation in the system. There is no such a "restarted" state so it can not be idempotent. We are discussing to implement in later operations a "wait" logic but for now it is just an operation.

Alex-Izquierdo:
I think there commits out of the scope of this PR. Can you rebase?

tests/integration/targets/activation/tasks/activation_lifecycle.yml
alinabuzachis: Then, I would suggest you add a note about the fact that the module is not idempotent when restarted.

Alex-Izquierdo (commented):
the pipeline is not passing: 
```
TASK [activation : Restart activation in check_mode] ***************************
fatal: [testhost]: FAILED! => {"changed": false, "msg": "Rulebook activation is not ready to be restarted. Status: starting."}
```

plugins/modules/rulebook_activation.py
```diff
@@ -53,6 +52,12 @@
     description:
       - The extra variables for the rulebook activation.
     type: str
+  restart:
+    description:
+      - Whether or not the activation should be restarted.
```
Alex-Izquierdo: I think this description can be clearer. I suggest something like:
```
Performs a restart of the activation. This is a non idempotent operation and if it is enabled the rest of parameters will be ignored. 
```

plugins/modules/rulebook_activation.py
```diff
@@ -539,6 +551,27 @@ def main() -> None:
         except EDAError as e:
             module.fail_json(msg=f"Failed to delete rulebook activation: {e}")
 
+    if restart:
+        try:
+            while activation["status"] not in ["running"]:
```
Alex-Izquierdo: This validation is business logic that replaces the backend logic and the module should not do it. If the activation can not be restarted is the backend who determines it. In fact already accepts restarts in other statuses.

plugins/modules/rulebook_activation.py
```diff
@@ -539,6 +551,27 @@ def main() -> None:
...
+            result = controller.restart_if_needed(
+                activation,
+                endpoint=f"activations/{activation['id']}/restart",
+            )
+            module.exit_json(**result)
+        except EDAError as e:
+            module.fail_json(
+                msg=(
+                    f"Failed to restart rulebook activation: {e} "
+                    f"Current Activation restart_count: {activation['restart_count']}"
```
Alex-Izquierdo, line 564: I don't think we have to arbitrarily return such a counter. Do you already have the rulebook_info module to get what you consider. Ithink it should return just the error returned by the backend.

plugins/modules/rulebook_activation.py
B-Whitt: updated

plugins/modules/rulebook_activation.py
B-Whitt: removed

zkayyali812 (approved):

zkayyali812 (approved):

bzwei (approved):

Alex-Izquierdo (approved):

PR #416, opened 2025-03-25
Move Activation Copy to Separate Module-AAP-42215

B-Whitt:
A new module is created for the copy action to copy the activation.

It has been discussed and decided that we shall move the existing copy action from the activation module to a separate module. The copy action shall take only one parameter which is the name of the copied activation. The resulting activation should be in disabled state. [https://issues.redhat.com/browse/AAP-42215]

### DIFF
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -16,7 +16,7 @@ author:
   - "Alina Buzachis (@alinabuzachis)"
 short_description: Manage rulebook activations in the EDA Controller
 description:
-  - This module allows the user to create or delete rulebook activations in the EDA Controller.
+  - This module allows the user to create, delete, or restart rulebook activations in the EDA Controller.
 options:
   name:
     description:
@@ -196,12 +196,7 @@ EXAMPLES = r"""
       - event_stream: "Example Event Stream"
         source_name: "Sample source"
 
-- name: Copy an existing rulebook activation
-  ansible.eda.rulebook_activation:
-    name: "Example Rulebook Activation - copy"
-    copy_from: "Example Rulebook Activation"
-
-- name: Rename a rulebook activation
+- name: Delete a rulebook activation
   ansible.eda.rulebook_activation:
     name: "Example Rulebook Activation"
     new_name: "Example Rulebook Activation New Name"
@@ -234,6 +229,7 @@ EXAMPLES = r"""
 - name: Restart activation
   ansible.eda.rulebook_activation:
     name: "Example Rulebook Activation - Restart"
+    organization_name: "Default"
     restart: true
 
 - name: Delete a rulebook activation
@@ -490,8 +486,6 @@ def create_params(
 def main() -> None:
     argument_spec = dict(
         name=dict(type="str", required=True),
-        new_name=dict(type="str"),
-        copy_from=dict(type="str"),
         description=dict(type="str"),
         project_name=dict(type="str", aliases=["project"]),
         rulebook_name=dict(type="str", aliases=["rulebook"]),
@@ -552,6 +546,20 @@ def main() -> None:
     copy_from = module.params.get("copy_from", None)
     restart = module.params.get("restart", None)
 
+    required_if = []
+    if not restart:
+        required_if = [
+            (
+                "state",
+                "present",
+                ("name", "rulebook_name", "decision_environment_name", "project_name"),
+            )
+        ]
+
+    module = AnsibleModule(
+        argument_spec=argument_spec, required_if=required_if, supports_check_mode=True
+    )
+
     if not HAS_YAML:
         module.fail_json(
             msg=missing_required_lib("pyyaml"), exception=YAML_IMPORT_ERROR
@@ -607,25 +615,8 @@ def main() -> None:
             )
             module.exit_json(**result)
         except EDAError as e:
-            module.fail_json(msg=(f"Failed to restart rulebook activation: {e}."))
+            module.fail_json(msg=f"Failed to restart rulebook activation: {e}.")
 
-    if copy_from:
-        try:
-            result = controller.copy_if_needed(
-                name,
-                copy_from,
-                endpoint=f"activations/{activation['id']}/copy",
-                item_type="activation",
-            )
-            module.exit_json(changed=True)
-        except KeyError as e:
-            module.fail_json(
-                msg=f"Unable to access {e} of the activation to copy from."
-            )
-        except EDAError as e:
-            module.fail_json(msg=f"Failed to copy rulebook activation: {e}")
-
-    # Parse credential IDs in existing activation
     if activation:
         credential_ids = [
             credential_id["id"] for credential_id in activation["eda_credentials"]
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -29,13 +29,6 @@ options:
       - If set, the rulebook activation will be updated with the new name.
     type: str
     version_added: 2.7.0
-  copy_from:
-    description:
-      - Name of the existing rulebook activation to copy.
-      - If set, copies the specified rulebook activation.
-      - The new rulebook activation will be created with the name given in the C(name) parameter.
-    type: str
-    version_added: 2.7.0
   description:
     description:
       - The description of the rulebook activation.
@@ -196,7 +189,7 @@ EXAMPLES = r"""
       - event_stream: "Example Event Stream"
         source_name: "Sample source"
 
-- name: Delete a rulebook activation
+- name: Rename a rulebook activation
   ansible.eda.rulebook_activation:
     name: "Example Rulebook Activation"
     new_name: "Example Rulebook Activation New Name"
@@ -486,6 +479,7 @@ def create_params(
 def main() -> None:
     argument_spec = dict(
         name=dict(type="str", required=True),
+        new_name=dict(type="str"),
         description=dict(type="str"),
         project_name=dict(type="str", aliases=["project"]),
         rulebook_name=dict(type="str", aliases=["rulebook"]),
@@ -541,11 +535,8 @@ def main() -> None:
         )
 
     argument_spec.update(AUTH_ARGSPEC)
-
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
-    copy_from = module.params.get("copy_from", None)
     restart = module.params.get("restart", None)
-
     required_if = []
     if not restart:
         required_if = [
@@ -594,9 +585,7 @@ def main() -> None:
     # Attempt to find rulebook activation based on the provided name
     activation = {}
     try:
-        activation = controller.get_exactly_one(
-            "activations", name=copy_from if copy_from else name
-        )
+        activation = controller.get_exactly_one("activations", name=name)
     except EDAError as e:
         module.fail_json(msg=f"Failed to get rulebook activation: {e}")
 
@@ -617,6 +606,7 @@ def main() -> None:
         except EDAError as e:
             module.fail_json(msg=f"Failed to restart rulebook activation: {e}.")
 
+    # Parse credential IDs in existing activation
     if activation:
         credential_ids = [
             credential_id["id"] for credential_id in activation["eda_credentials"]
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -537,19 +537,6 @@ def main() -> None:
     argument_spec.update(AUTH_ARGSPEC)
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
     restart = module.params.get("restart", None)
-    required_if = []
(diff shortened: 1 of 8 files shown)


plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+options:
+  name:
+    description:
+      - The name of the rulebook activation.
+    type: str
+    required: true
+  copy_from:
+    description:
+      - Name of the existing rulebook activation to copy.
+      - If set, copies the specified rulebook activation.
```
bzwei, line 26: copy_from is a required option, just like the name option

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+      - The new rulebook activation will be created with the name given in the C(name) parameter.
+    type: str
+    version_added: 2.6.0
+  description:
```
bzwei, line 30: All the following options should not exist

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+notes:
+
+"""
+
+EXAMPLES = r"""
+- name: Copy an existing rulebook activation
+  ansible.eda.rulebook_activation_copy:
+    name: "Example Rulebook Activation - copy"
+    copy_from: "Example Rulebook Activation"
+    organization_name: "Default"
```
bzwei, line 67: no organization_name argument

bzwei: Have you removed this example in rulebook_activation.py?

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+
+"""
+
+RETURN = r"""
+id:
```
bzwei, line 72: The api returns the whole activation object, not just the id. Have you verified the return format in the collection?

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+from ..module_utils.client import Client
+from ..module_utils.controller import Controller
+from ..module_utils.errors import EDAError
+
+
+def main() -> None:
+    argument_spec = dict(
+        name=dict(type="str", required=True),
+        copy_from=dict(type="str"),
+        description=dict(type="str"),
```
bzwei, line 92: no other arguments other than name and copy_from

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+        state=dict(choices=["present", "absent"], default="present"),
+    )
+
+    argument_spec.update(AUTH_ARGSPEC)
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    copy_from = module.params.get("copy_from", None)
+
+    required_if = []
+    if not copy_from:
```
bzwei, line 105: both name and copy_from are required. This required_if is incorrect

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+    client = Client(
+        host=module.params.get("controller_host"),
+        username=module.params.get("controller_username"),
+        password=module.params.get("controller_password"),
+        timeout=module.params.get("request_timeout"),
+        validate_certs=module.params.get("validate_certs"),
+    )
+
+    name = module.params.get("name")
+    state = module.params.get("state")
```
bzwei, line 127: no state

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+
+    controller = Controller(client, module)
+    # Organization is not available in Controller 2.4 API
+    config_endpoint_avail = controller.get_endpoint(
+        "config",
+    )
+    is_aap_24 = config_endpoint_avail.status in (404,)
+    organization_name = module.params.get("organization_name")
+    if state == "present" and not is_aap_24 and organization_name is None:
```
bzwei, line 136: no state check, or organzation_name check

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,171 @@
...
+    # Attempt to find rulebook activation based on the provided name
+    activation = {}
+    try:
+        activation = controller.get_exactly_one(
+            "activations", name=copy_from if copy_from else name
+        )
+    except EDAError as e:
+        module.fail_json(msg=f"Failed to get rulebook activation: {e}")
+
+    if copy_from:
```
bzwei, line 152: no if condition

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -211,14 +220,14 @@
...
+        organization_name: "{{ organization_name}}"
       register: _result
 
     - name: Copy rulebook activation
-      ansible.eda.rulebook_activation:
+      ansible.eda.rulebook_activation_copy:
         name: "{{ activation_name_copy }}"
         copy_from: "{{ activation_name }}"
-        organization_name: Default
+        organization_name: "{{ organization_name}}"
```
bzwei, line 230: no organization_name

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -256,10 +265,10 @@
           - _result.changed
 
     - name: Copy non-existing activation
-      ansible.eda.rulebook_activation:
+      ansible.eda.rulebook_activation_copy:
         name: "{{ activation_name_copy }}"
         copy_from: Invalid_activation
-        organization_name: Default
+        organization_name: "{{ organization_name}}"
```
bzwei, line 271: no organization_name

tests/integration/targets/activation/tasks/main.yml
```diff
@@ -270,15 +279,10 @@
...
+      ansible.eda.rulebook_activation_copy:
         name: "{{ activation_name_copy }}"
         copy_from: "{{ activation_name }}"
-        project_name: "{{ project_name }}"
-        rulebook_name: "{{ rulebook_name }}"
-        decision_environment_name: "{{ decision_env_name }}"
-        eda_credentials:
-          - "{{ credential_name_aap }}"
-        organization_name: Default
+        organization_name: "{{ organization_name}}"
```
bzwei, line 285: no organization_name

bzwei (changes requested):

plugins/modules/rulebook_activation_copy.py
Alex-Izquierdo: This is a thing that we should revisit at some point. Right not the collection only returns the id in all the module, delegating to its version "_info" module to fetch all the data. But I don't know if this is the standard, I see in other ansible modules that it returns the whole object. We might want to change this behavior for the whole collection but so far, returning the id would be inline with rest of the modules (if we are actually returning the id and not the whole object)

plugins/modules/rulebook_activation_copy.py
B-Whitt: updated

plugins/modules/rulebook_activation_copy.py
B-Whitt: organization_name is required when state is present. so I can remove from rulebook_activation_copy.py but not rulebook_activation.py

plugins/modules/rulebook_activation_copy.py
B-Whitt: removed

plugins/modules/rulebook_activation_copy.py
B-Whitt: removed

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    name = module.params.get("name")
+    controller = Controller(client, module)
+
+    # Attempt to find rulebook activation based on the provided name
+    activation = {}
+    try:
+        activation = controller.get_exactly_one(
+            "activations", name=copy_from if copy_from else name
```
bzwei, line 88: copy_from is mandatory so it should be just `name=copy_from` without if and else

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,108 @@
...
+def main() -> None:
+    argument_spec = dict(
+        name=dict(type="str", required=True),
+        copy_from=dict(type="str", required=True),
+    )
+
+    argument_spec.update(AUTH_ARGSPEC)
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    copy_from = module.params.get("copy_from", None)
```
bzwei, line 66: module.params.get("copy_from")

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,108 @@
...
+options:
+  name:
+    description:
+      - The name of the new rulebook activation.
+    type: str
+    required: true
+  copy_from:
+    description:
+      - Name of the existing rulebook activation to copy.
+      - If set, copies the specified rulebook activation.
```
bzwei, line 26: This line is not needed

B-Whitt: removed

plugins/modules/rulebook_activation_copy.py
B-Whitt: updated

plugins/modules/rulebook_activation_copy.py
B-Whitt: updated

bzwei (approved):

tests/run-staging
```diff
@@ -25,4 +25,4 @@ ansible-test coverage report --requirements --omit '.tox/*,tests/*' --color --al
 ansible-test coverage combine "--export=${TOX_ENV_DIR:-.}"
 
 # Clean up containers
-bash -c 'cd ${TOX_WORK_DIR}/eda-server/tools/docker && docker compose -p eda -f docker-compose-stage.yaml down'
+bash -c 'cd ${TOX_WORK_DIR}/eda-server/tools/docker && docker compose -p eda -f docker-compose-stage.yaml --profile proxy down'
```
zkayyali812, line 28: Is this change done on purpose?

B-Whitt: yes. that was merged in from Kaio's PR (#417)

zkayyali812 (dismissed):

plugins/modules/rulebook_activation.py
```diff
@@ -547,10 +535,21 @@ def endpoint_state(activation: Dict[str, Any], state: str) -> str:
         )
 
     argument_spec.update(AUTH_ARGSPEC)
-
     module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
-    copy_from = module.params.get("copy_from", None)
     restart = module.params.get("restart", None)
+    required_if = []
```
kaiokmo, line 540: @B-Whitt Why we are reintroducing the `required_if` parameter? It was removed to be able to enable/disable activations with only providing name and state, and the logic changed to let the backend return in case of missing fields.

Maybe you probably need to rebase from main first?

kaiokmo (changes requested):

plugins/modules/rulebook_activation_copy.py
```diff
@@ -0,0 +1,109 @@
...
+    description:
+      - The name of the new rulebook activation.
+    type: str
+    required: true
+  copy_from:
+    description:
+      - Name of the existing rulebook activation to copy.
+    type: str
+    required: true
+    version_added: 2.6.0
```
kaiokmo, line 31: I think this option is not supposed to exist in this context, since this is a brand-new module. Instead, I believe we need to include a version_added field in the top-level documentation. Check https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-fields:~:text=Generally%20not%20needed.-,version_added,-%3A

kaiokmo (changes requested):

plugins/modules/rulebook_activation_copy.py
B-Whitt: moved version_added

bzwei (approved):

kaiokmo:
@B-Whitt are we planning to remove the existing copy functionality from the rulebook_activation module in a different PR, and also change the integration tests accordingly?

B-Whitt:
> @B-Whitt are we planning to remove the existing copy functionality from the rulebook_activation module in a different PR, and also change the integration tests accordingly?

@kaiokmo `copy_from` was removed from `rulebook_activation` in this PR. rulebook_activation_copy integration tests are still with the rulebook_activation integration tests as they are with other modules. We can break out integration tests for each module in a later PR.

zkayyali812 (approved):

kaiokmo (approved):

PR #421, opened 2025-04-11
fix: pass in optional arguments to sqs

mkanoor:
Based on feedback we received we are missing some key parameters from sqs queue processing. Adding additional optional parameters

* queue_owner_aws_account_id
* max_number_of_messages

https://issues.redhat.com/browse/AAP-42623

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/get_queue_url.html

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/receive_message.html

### DIFF
--- a/extensions/eda/plugins/event_source/aws_sqs_queue.py
+++ b/extensions/eda/plugins/event_source/aws_sqs_queue.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import logging
-from typing import Any
+from typing import Any, TypedDict
 
 import botocore.exceptions
 from aiobotocore.session import get_session
@@ -31,6 +31,13 @@ options:
     description:
       - Optional URL to connect to instead of the default AWS endpoints.
     type: str
+  queue_owner_aws_account_id:
+    description:
+      - >
+        Optional The Amazon Web Services account ID of the account that
+        created the queue. This is only required when you are attempting
+        to access a queue owned by another Amazon Web Services account.
+    type: str
   region:
     description:
       - Optional AWS region to use.
@@ -46,6 +53,14 @@ options:
       - Set to 0 to disable.
     type: int
     default: 2
+  max_number_of_messages:
+    description:
+      - >
+        Optional The maximum number of messages to return. Amazon SQS
+        never returns more messages than this value (however, fewer
+        messages might be returned). Valid values are 1 to 10.
+    type: int
+    default: 1
 """
 
 EXAMPLES = r"""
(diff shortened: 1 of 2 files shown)


hsong-rh (approved):

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -46,6 +52,14 @@
       - Set to 0 to disable.
     type: int
     default: 2
+  max_number_of_messages:
+    description:
+      - Optional: The maximum number of messages to return. Amazon SQS
```
bzwei, line 57: Comparing with other uses of Optional, it seems `:` is not needed.

bzwei (dismissed):

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -31,6 +31,12 @@
     description:
       - Optional URL to connect to instead of the default AWS endpoints.
     type: str
+  queue_owner_aws_account_id:
+    description:
+      - Optional The Amazon Web Services account ID of the account that
+      - created the queue. This is only required when you are attempting
```
Dostonbek1, line 37: Adding dash (`-`) creates a new paragraph in documentation. So we should just have one on top and remove the rest, unless we mean to add a paragraph.

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -46,6 +52,14 @@
       - Set to 0 to disable.
     type: int
     default: 2
+  max_number_of_messages:
+    description:
+      - Optional The maximum number of messages to return. Amazon SQS
+      - never returns more messages than this value (however, fewer
```
Dostonbek1, line 58: Same here regarding dash `-` here as above.

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -46,6 +52,14 @@
...
+      - messages might be returned). Valid values: 1 to 10.
+      - Default: 1.
```
Dostonbek1, line 60: No need to add `Default` here as it is defined as a separate param below. Refer to [this](https://console.redhat.com/ansible/automation-hub/repo/published/ansible/eda/content/eda%2Fplugins%2Fevent_source/alertmanager/) to see how it is displayed.

Dostonbek1 (changes requested):

Dostonbek1 (approved):

hsong-rh (approved):

bzwei (approved):

PR #428, opened 2025-05-06
fix: remove duplicates and update mismatched python versions in tox.yml

Dostonbek1:
- remove duplicate `py312-unit` in tox.yml
- change mismatched py312 with py311


ptoscano (approved):
Nice consistency fix :+1:

ttuffin (approved):

Issue #401, opened 2025-02-26
azure_service_bus raises  QueueFull error


susanhooks:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/ansible-rulebook/issues) for duplicates.
- [x] I understand that event-driven-ansible collection is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When triggering a large number of events, QueueFull error is raised, rather than the expected behavior of waiting until there's space in the queue. 

I believe the asyncio `put` method as opposed to `put_nowait` should be implemented, as was done in  #342 

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.9/site-packages/ansible_rulebook/engine.py", line 97, in put_nowait
    self.queue.put_nowait(data)
  File "/usr/local/lib/python3.9/asyncio/queues.py", line 151, in put_nowait
    raise QueueFull
asyncio.queues.QueueFull
```

### Environment

```
root@403bcda8c6b9:/local# ansible-rulebook --version
1.1.2
  Executable location = /usr/local/bin/ansible-rulebook
  Drools_jpy version = 0.3.9
  Java home = /usr/lib/jvm/java-17-openjdk-amd64
  Java version = 17.0.14
  Python version = 3.9.21 (main, Feb 25 2025, 05:18:57) [GCC 12.2.0]

root@403bcda8c6b9:/local# ansible-galaxy collection list

# /root/.ansible/collections/ansible_collections
Collection                             Version
-------------------------------------- -------
ansible.eda                            2.5.0
ansible.utils                          5.1.2
ansible.windows                        2.7.0
chevron_eda.nautobot_eda_configuration 1.0.3
cisco.ise                              2.10.0
community.windows                      2.3.0
delinea.ss                             1.1.0
kentik.kentik_config                   1.1.1
networktocode.nautobot                 5.8.0
servicenow.servicenow                  1.0.6
```

### Steps to reproduce

Create a rulebook connected to azure_service_bus as a message bus, trigger a large number of events. In this case, we were triggering update to DNS via JobHook in Nautobot. 

Sample event below. 

```
[...]

PR #451, opened 2025-07-01
Azure servicebus - add async functionality

susanhooks:
This PR is to fix the issue #401 , in which the async functionality was not working as expected. I've rewritten this module to leverage the Azure Service Bus async functionality. 

In my testing, this resolves the QueueFull message that was being received when a large number of events needed to be processed.

### DIFF
--- a/extensions/eda/plugins/event_source/azure_service_bus.py
+++ b/extensions/eda/plugins/event_source/azure_service_bus.py
@@ -1,10 +1,9 @@
 import asyncio
-import concurrent.futures
 import contextlib
 import json
 from typing import Any
 
-from azure.servicebus import ServiceBusClient
+from azure.servicebus.aio import ServiceBusClient
 
 DOCUMENTATION = r"""
 ---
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
Hi @susanhooks 
Thank you very much for you contribution. The PR looks good but as you can see in the CI, your changes need also to update the tests.

susanhooks:
> Hi @susanhooks Thank you very much for you contribution. The PR looks good but as you can see in the CI, your changes need also to update the tests.

Thank you, I will update tests for this.

Alex-Izquierdo (approved):

ptoscano:
> Would you mind re-submiting this PR to that collection? When I did the initial migration I was careful to keep all the git history so I would really like your commit to come from you.

Let's keep this process light on contributors (be them Red Hat folks or external) for now; I created a PR forwarding the commit as-is of this change: https://github.com/ansible-collections/azure/pull/2017

PR #455, opened 2025-07-29
Deprecate azure_service_bus

p3ck:
azure_service_bus plugin has been moved to
azure.azcollection.azure_service_bus and will be removed in a future release.

This is my first time deprecating a module that has moved so I may have missed something.  Thank you

### DIFF
--- a/extensions/eda/rulebooks/github-ci-cd-rules.yml
+++ b/extensions/eda/rulebooks/github-ci-cd-rules.yml
@@ -1,7 +1,7 @@
 - name: GitHub Deploy Git Hook Rules
   hosts: all
   sources:
-    - ansible.eda.azure_service_bus:
+    - azure.azcollection.azure_service_bus:
         conn_str: "{{connection_str}}"
         queue_name: "{{queue_name}}"
       filters:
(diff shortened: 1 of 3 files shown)


ptoscano (changes requested):
Hi @p3ck,

thanks for approaching us about the Azure EDA event source.

I noted inline why the current PR cannot work: as also @Alex-Izquierdo wrote, there is no support in ansible-rulebook/eda-server for any of the module lifecycle feature that ansible-playbook supports. I think what only works is marking the EDA module as deprecated in its `DOCUMENTATION` block.

p3ck:
Hello,

I was on PTO for the last few weeks so just getting back to this now.  It doesn't seem like any of my changes to deprecate the module from here will work.  Do you have any suggestion on how to proceed?

Thanks,
Bill

Alex-Izquierdo:
> Hello,
> 
> I was on PTO for the last few weeks so just getting back to this now. It doesn't seem like any of my changes to deprecate the module from here will work. Do you have any suggestion on how to proceed?
> 
> Thanks, Bill

@p3ck I think that, as a short-term solution, we are going to deprecate it through documentation only. Enabling such deprecation behavior in ansible-rulebook would require a bigger effort, and I believe it’s not part of our short-term plans.

What we can also do is print a [deprecation warning message](https://docs.python.org/3/library/warnings.html) at the start of the module.

PR #457, opened 2025-07-31
feat: support subscription to multiple topics or by a pattern

bzwei:


extensions/eda/plugins/event_source/kafka.py
```diff
@@ -61,7 +61,16 @@
...
+      - The kafka topic. topic, topics, and topic_pattern are mutually exclusive.
+    type: str
+  topics:
+    description:
+      - The kafka topics. topic, topics, and topic_pattern are mutually exclusive.
+    type: list
+    elements: str
+  topic_pattern:
+    description:
+      - The kafka topic pattern. It must be a valid regex. topic, topics, and topic_pattern are mutually exclusive.
```
mkanoor, line 73: @bzwei Can this be multiple regex patterns or just one?

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -129,6 +140,17 @@ async def main(  # pylint: disable=R0914
     logger = logging.getLogger()
 
     topic = args.get("topic")
+    topics = args.get("topics")
+    topic_pattern = args.get("topic_pattern")
```
mkanoor, line 144: @bzwei It might make sense to call re compile on the string to see if its a valid regex.
```
import re
try:
        re.compile(topic_pattern)
except re.error:
       raise ValueError("Invalid regex " + topic_pattern + " error: str(re.error))
```

Alex-Izquierdo (dismissed):

PR #476, opened 2025-09-19
feat: Add support for multiple Kafka brokers

sevenup30:
This PR enhances the Kafka consumer configuration by adding support for connecting to multiple brokers instead of a single one.

In most non-local environments, Kafka clusters run with at least three brokers for high availability. This change allows our consumer to be aware of the entire cluster, as supported by the underlying AIOKafkaConsumer library.

Additionally, this PR introduces the client_id setting. This is a best practice for identifying and tracking the consumer's activity on the Kafka brokers, which is valuable for monitoring and debugging purposes.

Reference: [AIOKafkaConsumer API Documentation](https://aiokafka.readthedocs.io/en/stable/api.html#aiokafka.AIOKafkaConsumer)

### DIFF
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -166,7 +166,6 @@ def _host_or_broker_validation(host: str, port: int, brokers: list[str]) -> None
         raise ValueError(msg)
 
 
-
 async def main(  # pylint: disable=R0914
     queue: asyncio.Queue[Any],
     args: dict[str, Any],
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 import logging
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
-from typing import Any
+from typing import Any, Optional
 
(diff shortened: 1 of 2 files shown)


ptoscano:
Hi @sevenup30,

first of all, thanks for the PR!

On the actual code:
- replacing the `host` and `port` parameters is a breaking change, and I'm not sure it is worth at this time (every usage of it by users would need to be changed); IMHO a backward-compatible addition would be better;
   - a potential idea could be to add `servers` as list of servers, and make it mutually exclusive with `host`+`port` (i.e. either specify the existing keys or the new one)
- a comma-separated list as string does not seem very yaml-y; please make it as a proper list, so there is less need to deal with item splitting (in the source) or joining (by users)
- the `client_id` addition seems fine; what do you think about splitting it in its own PR, so it can be easily reviewed and merged without waiting for the discussion on the API for multiple servers?

Thanks!

paspflue:
@sevenup30 what do you think about renaming `brokers` to `bootstrap_servers`? That would make it similar to the Kafka namings.

PS: I would also be interested in this feature.

paspflue:
@ptoscano from your perspective, is there anything else that needs to be addressed here?
I'm interested in this feature and have already tested it locally.

The parameter name “brokers” can also ramain.


Is there any other reason not to include this?

PR #495, opened 2025-12-11
feat: add docstrings and enable pylint docstring checks

DenisKoleda:
## Summary
- Enable missing-module-docstring, missing-class-docstring, missing-function-docstring checks in pyproject.toml
- Add docstrings to all modules, event_source and event_filter plugins
- Add docstrings to module_utils (client, controller, common, errors, arguments)
- Add docstrings to docs/conf.py and doc_fragments/eda_controller.py
- Replace walrus operator with standard assignment in tests
- Add 'rtype' to cspell dictionary

## Test plan
- [x] pre-commit passes
- [x] All files compile successfully

## Summary by CodeRabbit

* **New Features**
  * Added Kafka event-source support for Avro message deserialization, including schema registries, authentication, SSL configuration, retries, and related metadata handling.

* **Documentation**
  * Added comprehensive module- and function-level documentation across event sources, filters, Ansible modules, and utility helpers.

* **Chores**
  * Updated linting configuration to enforce module, class, and function documentation and added supporting static-analysis configuration.


ptoscano (changes requested):
Hi @DenisKoleda,

thanks for the contribution! Since it is more than 1000 lines, please give us a bit of time to review it, and provide more feedback.

I did a quick review, and noted few things to fix inline; in addition to that:

>     * Replace walrus operator with standard assignment in tests

What is the reason for this change? Since this is a code change rather than a pure documentation change like the rest of the bits in this PR, it needs an explanation. In any case, if it is really needed, I'd do it in a separate PR, so it can be reviewed on its own.

What is the motivation behind these changes? From a quick skimming I see there is a potential redundancy of doctext/descriptions, especially for modules and their `main()` functions.

I'm also not convinced about the manual specification of `:type` and `:rtype`: the codebase use type hints, so maintaining the same details on types of parameters and return values adds duplication. It would be better if sphinx was able to use the type hints automatically; have you checked whether it is possible?

Thanks!

DenisKoleda:
Hi @ptoscano,

Thank you for the review! I've addressed all the feedback in the latest commit:

1. **Walrus operator changes** - Removed from this PR. Will submit as a separate PR.

2. **docs/conf.py** - Reverted the docstring. Added `# pylint: disable=missing-module-docstring` instead, since the linter was complaining after enabling docstring checks.

3. **:type and :rtype** - Removed from all docstrings (39 files).

4. **Module/main() redundancy** - Simplified all main() docstrings to "Module entry point." since module-level docstrings already describe the purpose.

Regarding sphinx-autodoc-typehints: I checked this - for Ansible collections, user-facing documentation is generated from DOCUMENTATION YAML blocks by antsibull-docs, not from Python docstrings. The Python docstrings serve as internal developer documentation only. I verified the Sphinx build completes successfully with these changes.

Please let me know if you'd like any further changes!

ptoscano:
I was out on vacation, hence sorry for the late answer.
 
Thanks for the updates: they solve some of the concerns/notes I had. That said, there is still a bigger concern I have:

> What is the motivation behind these changes?

In particular, I don't understand what would be the gained value of documenting the Python of the plugins (both Ansible and EDA ones), especially when the modules exist only to implement the plugins (and thus they have their own documentation as `DOCUMENTATION` block). In practice the documentation of each plugin is also the "documentation" for its Python module, and details on the internals better fit as inline code comments.

I still see duplication of `DOCUMENTATION` and the added docstring for EDA plugins. I'd simply not change the docstrings of `main()` at all, and improve `DOCUMENTATION` instead (e.g. w.r.t. the parameters).

Also, documenting the returned exceptions of internal functions may risk to never be comprehensive; a couple of random examples:
- `extensions/eda/plugins/event_source/file.py`, function `send_facts()`: the added docstring mentions `TypeError`, however it can also raise other exceptions due to the usage of `pathlib` (say a specified file exists, but it is not readable by the current user)
- `extensions/eda/plugins/event_source/pg_listener.py, function `_validate_args()`: it can raise `MissingRequiredArgumentError`, which is not a `ValueError`
- `extensions/eda/plugins/event_source/webhook.py`, function `_get_ssl_context()`: it can re-raise exception raisen by functions of `ssl.SSLContext`

I hope I'm not looking dismissive here, however I'm not getting the "whole picture" of this work; would you please detail a bit more what is the "end goal" you are trying to achieve? Say be able to use some tool (IDE, etc) on the codebase, or consume the documentation in some form (which?), etc.

Thanks again for your interest!

ttuffin (dismissed):
Thanks for the contribution!

ttuffin (dismissed):

kaiokmo (changes requested):
Hey @DenisKoleda! Let's get this one merged.

Can you please rebase your branch on top of main and resolve the conflict on the kafka event source plugin? We should be ok after it.

Thanks!

DenisKoleda:
Hi @kaiokmo, done! I've rebased the branch onto the latest `main` and resolved the conflict in the Kafka event source plugin, preserving the current upstream implementation and the docstring changes from this PR. The updated branch has been pushed, and GitHub reports no merge conflicts.

All 113 Kafka unit tests passed, along with the pre-commit checks for the changed files (mypy was skipped locally because Python 3.9 is not available). Thanks for following up!

PR #11, opened 2022-08-16
Add ansible-test unit test for webhook plugin

hsong-rh:
Passed test command:
```
ansible-test units --docker default
```

PR #528, opened 2026-03-26
feat(kafka): add Avro deserialization and Schema Registry support

B-Whitt:
## Summary

- Add Avro deserialization to the Kafka event source plugin via `fastavro`
  with three modes: local `.avsc` schema files, Avro Object Container
  (self-describing), and Confluent-compatible Schema Registry lookup
- Implement Schema Registry integration with wire format auto-detection,
  bounded LRU schema cache (100 entries), and three auth methods
  (Basic Auth, Bearer Token, OAuth 2.0 Client Credentials)
- Integrate upstream UUID validation (PR #545): every event gets a valid
  RFC 4122 UUID under `meta.uuid` — deterministic UUID5 from Kafka
  coordinates by default, overridable via `message_uuid` header or body
  field, with fallback for invalid UUIDs
- Add `meta.source_offset` (`{topic}:{partition}:{offset}`) for tracing
  messages back to their Kafka coordinates
- Skip Kafka tombstone messages (null values) gracefully
- Refactor message decoding into `_decode_message_body()` supporting
  both JSON and Avro formats
- Zero new runtime dependencies — `fastavro` is optional (only needed for
  Avro), `aiohttp` is already a transitive dep of `aiokafka`
- Add 8 new plugin parameters for Schema Registry configuration
- Compatible with Confluent, Karapace, Redpanda, and Apicurio registries
- Fully backward compatible — existing JSON configurations are unaffected

## Test plan

- [x] Run all 70+ unit tests: `tox -e unit -- tests/unit/event_source/test_kafka.py -v`
- [x] Verify JSON backward compatibility: `-k "test_receive_from_kafka or test_json_format_unchanged"`
- [x] Verify UUID validation: `-k "test_message_uuid or test_generated_uuid or test_invalid_uuid"`
- [x] Verify Avro with local schema: `-k "TestAvroDeserializer"`
- [x] Verify Avro integration through main: `-k "TestAvroKafkaIntegration"`
- [x] Verify Schema Registry config validation: `-k "TestSchemaRegistryConfig"`
- [x] Verify wire format detection: `-k "TestWireFormatDetection"`
- [x] Verify schema cache eviction: `-k "TestSchemaCache"`
- [x] Verify Schema Registry deserialization: `-k "TestSchemaRegistryDeserialize"`
- [x] Verify auth header generation: `-k "TestGetAuthHeaders"`
- [x] Verify OAuth token fetching: `-k "TestFetchOAuthToken"`
- [x] Verify SSL context handling: `-k "TestSSLContext"`
- [x] Verify tombstone handling: `-k "test_tombstone or test_decode_message_body_returns_none"`
- [x] Verify deserializer cleanup: `-k "TestDeserializerCloseViaMain"`
- [x] Ruff lint: `tox -e ruff`
- [x] End-to-end test with live Kafka + Schema Registry (see integration test README)

Assisted by: Claude Code


extensions/eda/plugins/event_source/kafka.py
```diff
@@ -166,12 +244,405 @@
...
+                async with session.post(self._oauth_token_url, data=data) as resp:
+                    resp.raise_for_status()
+                    token_data = await resp.json()
+                    self._oauth_access_token = token_data["access_token"]
+                    self._oauth_token_expiry = (
+                        time.time() + token_data.get("expires_in", 3600) - 30
+                    )
+                    self._logger.info("OAuth token obtained successfully")
+        except Exception:  # pylint: disable=broad-exception-caught
+            self._logger.exception("Failed to obtain OAuth token")
```
AlexSCorey, line 486: If we fail here we are likely not authenticated, should the process stop and bubble the errror up?  I worry that if it doesn't stop we will just boil over with other failures because we aren't authenticated.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -166,12 +244,405 @@
...
+        try:
+            async with aiohttp.ClientSession() as session:
+                data = {
+                    "grant_type": "client_credentials",
+                    "client_id": self._oauth_client_id,
+                    "client_secret": self._oauth_client_secret,
+                }
+                if self._oauth_scope:
+                    data["scope"] = self._oauth_scope
+                async with session.post(self._oauth_token_url, data=data) as resp:
```
AlexSCorey:
```suggestion
                async with session.post(self._oauth_token_url, data=data, ssl=self._ssl_context) as resp:
```
This will allow us to reuse the already that was built from the Kafka SSL settings when `schema_registry_ssl` is enabled..  

Without it, the token request uses aiohttp’s default TLS settings, which may not trust the same private CA and won’t present the same client cert—so it can fail in environments requiring custom CA or mTLS.

AlexSCorey (commented):
Just some suggestions.  Its probably ok without these suggestions, but they might make it more durable.

tests/integration/event_source_kafka/test_kafka_rules_avro.yml
```diff
@@ -0,0 +1,21 @@
+- name: test kafka source plugin with avro message format
+  hosts: localhost
+  sources:
+    - ansible.eda.kafka:
+        topic: kafka-events-avro
+        host: localhost
+        port: 9092
+        offset: earliest
+        message_format: avro
+        avro_schema_file: AVRO_SCHEMA_FILE_PLACEHOLDER
```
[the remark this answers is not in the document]
B-Whitt: added avro_schema_file: "{{ avro_schema_file }}"

extensions/eda/plugins/event_source/schemas/kafka.json
```diff
@@ -117,6 +117,65 @@
...
+            "title": "Message Format",
+            "description": "The deserialization format for Kafka message values. Use 'json' for JSON messages (default) or 'avro' for Apache Avro binary messages.",
+            "type": "string",
+            "enum": [
+                "json",
+                "avro"
+            ],
+            "default": "json"
+        },
+        "avro_schema_file": {
```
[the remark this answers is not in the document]
B-Whitt: I will add the sample gist as we discussed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: resolved

extensions/eda/plugins/event_source/kafka.py
B-Whitt: added... 
- First token failure → raises RuntimeError with a clear message. This stops the plugin immediately on misconfiguration.
- Refresh failure (existing token) → logs a warning with time remaining on the stale token, keeps running.

B-Whitt:
@coderabbitai review

extensions/eda/plugins/event_source/_avro.py
```diff
@@ -0,0 +1,420 @@
...
+        # Build SSL context
+        if use_ssl and parsed.scheme == "https":
+            self._ssl_context = ssl.create_default_context()
+            if ssl_cafile:
+                self._ssl_context.load_verify_locations(ssl_cafile)
+            if ssl_certfile and ssl_keyfile:
+                self._ssl_context.load_cert_chain(
+                    certfile=ssl_certfile,
+                    keyfile=ssl_keyfile,
+                )
```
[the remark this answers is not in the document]
B-Whitt: outdated

extensions/eda/plugins/event_source/_avro.py
```diff
@@ -0,0 +1,420 @@
...
+                    registry_schema,
+                )
+            self._logger.warning(
+                "Registry lookup failed for schema ID %d, trying fallbacks",
+                schema_id,
+            )
+
+        # Attempt 2: Schemaless deserialization with local schema
+        if self._schema is not None:
+            return self._deserialize_with_schema(data, self._schema)
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

extensions/eda/plugins/event_source/schemas/kafka.json
```diff
@@ -129,6 +129,65 @@
...
+        "schema_registry_oauth_token_url": {
+            "title": "Schema Registry OAuth Token URL",
+            "description": "OAuth 2.0 token endpoint URL for obtaining access tokens.",
+            "type": "string"
+        },
+        "schema_registry_oauth_scope": {
+            "title": "Schema Registry OAuth Scope",
+            "description": "OAuth 2.0 scope to request when obtaining access tokens.",
+            "type": "string"
+        },
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

tests/integration/event_source_kafka/test_kafka_source.py
```diff
@@ -1,10 +1,14 @@
...
+import io
 import json
 import os
 import platform
 import re
 import subprocess
 from typing import Any, Generator, Optional
 
+import fastavro
 import pytest
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

tests/unit/event_source/test_kafka.py
```diff
@@ -1,14 +1,24 @@
...
 import json
+import struct
+import time
 from typing import Any
+from unittest.mock import AsyncMock as StdAsyncMock
 from unittest.mock import MagicMock, patch
 
+import aiohttp
+import fastavro
 import pytest
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

tests/unit/event_source/test_kafka.py
```diff
@@ -752,3 +762,999 @@ async def mock_wait_for(coroutine: Any, timeout: float) -> Any:
...
+                self.index = 0
+
+            async def __anext__(self) -> MagicMock:
+                if self.index < len(self.messages):
+                    mock = MagicMock()
+                    mock.value = self.messages[self.index]
+                    mock.headers = [("source", b"test")]
+                    self.index += 1
+                    return mock
+                raise StopAsyncIteration
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -166,12 +254,436 @@
...
+        schema_registry_bearer_token: str | None = None,
+        schema_registry_oauth_client_id: str | None = None,
+        schema_registry_oauth_client_secret: str | None = None,
+        schema_registry_oauth_token_url: str | None = None,
+        schema_registry_oauth_scope: str | None = None,
+        schema_registry_ssl: bool = True,
+        ssl_cafile: str | None = None,
+        ssl_certfile: str | None = None,
+        ssl_keyfile: str | None = None,
+    ) -> None:
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -166,12 +254,436 @@
...
+                schema_str = resp_data["schema"]
+                # cast: fastavro lacks type stubs (mypy no-any-return)
+                schema = fastavro.parse_schema(json.loads(schema_str))
+                parsed_schema = cast("dict[str, Any]", schema)
+        except Exception:  # pylint: disable=broad-exception-caught
+            self._logger.exception(
+                "Failed to fetch schema ID %d from registry",
+                schema_id,
+            )
+            return None
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -395,14 +944,16 @@ async def receive_msg(
...
     """Receive messages from the Kafka topic and put them into the queue."""
     logger = logging.getLogger()
 
     async for msg in kafka_consumer:
         event: dict[str, Any] = {
             "meta": {
-                "uuid": f"{msg.topic}:{msg.partition}:{msg.offset}",
+                "source_offset": f"{msg.topic}:{msg.partition}:{msg.offset}",
                 "produced_at": get_message_timestamp(msg),
             },
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

tests/integration/event_source_kafka/README.md
```diff
@@ -0,0 +1,207 @@
...
+```yaml
+# avro-test-rulebook.yml
+---
+- name: Avro Test
+  hosts: localhost
+  sources:
+    - ansible.eda.kafka:
+        host: localhost
+        port: "9092"
+        topic: avro-alerts
```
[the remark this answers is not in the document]
B-Whitt: addressed. see comments below.

B-Whitt:
## Review Response — CodeRabbit & Human Reviewer Findings

All CodeRabbit and human reviewer comments have been investigated. Below is the resolution for each item. The latest commit (`7a7738b`) addresses all actionable findings.

---

### Resolved Issues (Fixed in Latest Commit)

#### 1. CI: pylint `disallowed-name` failure
CodeRabbit flagged docstring coverage at 46.15% < 80%, but that threshold is a CodeRabbit advisory (its own status was SUCCESS). The actual CI failure was a **pylint error**: `disallowed-name "_"` in `test_kafka.py:1730`. Renamed `_` to `_msg_format`. All functions already have docstrings.

#### 2. CRITICAL: Schema Registry fetch swallows ALL exceptions silently (item #4)
**Valid finding.** The broad `except Exception` at `_get_schema_from_registry()` made transient failures (registry down, auth expired, network timeout) indistinguishable from permanent ones (schema not found). With `auto_commit=True`, silently skipped messages are committed and lost.

**Fix:** Split exception handling to match Confluent client behavior:
- `ClientResponseError` with 404 → returns `None` (permanent, schema not found)
- `ClientResponseError` with other status codes (401, 500, etc.) → re-raises (transient)
- `ClientError` / `asyncio.TimeoutError` → re-raises (transient, network)
- `KeyError` / `json.JSONDecodeError` → returns `None` (permanent, malformed response)

Caller (`deserialize()`) catches transient exceptions and falls back to local schema if available, otherwise re-raises to the consumer loop. **4 new tests** cover all paths (94 total, all passing).

#### 3. MAJOR: SSL setup never forwards `password` to `load_cert_chain()` (item #6)
**Valid finding.** The Schema Registry SSL path skipped the `password` parameter entirely, so encrypted (passphrase-protected PEM) keyfiles would fail with `ssl.SSLError`.

**Fix:** Threaded `ssl_password` through the full chain: `main()` → `_configure_avro()` → `AvroDeserializer.__init__()` → `_configure_registry()` → `load_cert_chain(password=ssl_password)`. No behavior change for unencrypted keys (`password=None` is ignored).

#### 4. MINOR: Mock message missing attributes in unit test (item #10)
**Valid finding.** `AvroAsyncIterator` was missing `timestamp`, `topic`, `partition`, `offset` — producing MagicMock objects that got stringified into garbage in `meta.source_offset`. Added the missing attributes and `source_offset` assertions matching the pattern of all other iterators.

---

### Investigated — Not Bugs (No Action Needed)

#### 5. CRITICAL: Wire-format fallback passes full bytes to local schema (item #3)
CodeRabbit referenced `_avro.py:325-341` — **a file from an intermediate commit that no longer exists.** The current code in `kafka.py` explicitly strips the 5-byte header (`data[self._WIRE_FORMAT_HEADER_SIZE :]`) before falling back to local schema. The test `test_wire_format_registry_failure_falls_back_to_local` validates this path with a mocked 404 failure and confirms correct deserialization.

#### 6. CRITICAL: Event dict missing fallback `meta.uuid` (item #5)
CodeRabbit compared against the old broken code, not upstream `main`. Upstream `main` **never sets `meta.uuid`** — the `insert_meta_info` event filter generates a UUID4 when absent. The old pre-PR code set `meta.uuid = f"{topic}:{partition}:{offset}"` (a non-UUID string that broke eda-server's `AuditEvent(id=UUIDField)`, causing silent audit event loss). The PR correctly renamed the coordinate to `meta.source_offset` and leaves `meta.uuid` unset for the filter to populate.

#### 7. MAJOR: `requirements.txt` — fastavro listed as hard dep with "optional" comment (item #7)
CodeRabbit referenced an earlier commit. Current `requirements.txt` has no "optional" comment on the `fastavro` line — it follows the same flat-list pattern as all other plugin-specific deps (`aiobotocore`, `azure-servicebus`, `psycopg`, `systemd-python`). The `try/except ImportError` + `HAS_FASTAVRO` flag is a runtime safety net, consistent with every other plugin in the collection.

#### 8. MAJOR: JSON schema should encode auth mutual exclusion (item #8)
The JSON schema files in `schemas/` are **not consumed by any component** in the current stack (eda-server doesn't read them, ansible-rulebook doesn't validate against them, EDA Controller UI doesn't generate forms from them). No other schema in the collection uses `oneOf`/`anyOf`/`if-then`. Auth mutual exclusion IS enforced in Python code (`kafka.py:408-415`) and tested by `TestSchemaRegistryConfig::test_auth_mutual_exclusion`.

#### 9. MAJOR: Top-level `import fastavro` in test files (item #9)
`fastavro` is in all three requirements files and always installed in every CI/tox environment. Every other test file uses top-level imports the same way (`test_pg_listener.py` → `psycopg`/`xxhash`, `test_azure_service_bus.py` → Azure types). No test file in the collection uses `pytest.importorskip()`.

#### 10. MINOR: Port should be integer in README example (item #11)
The plugin's own `DOCUMENTATION` declares `port` as `type: str` and all four plugin EXAMPLES use `port: "9092"`. The README matches this convention.

---

### Human Reviewer Comments (All Resolved)

| Reviewer | Issue | Resolution |
|----------|-------|------------|
| **@AlexSCorey** | OAuth first-failure should stop, not continue silently | First failure → `RuntimeError`; refresh failure → warning with time remaining |
| **@AlexSCorey** | OAuth token request should use `ssl=self._ssl_context` | SSL context reuse added for token endpoint |
| **@mkanoor** | Rulebook should use Jinja substitution for credential params | Added `avro_schema_file: "{{ avro_schema_file }}"` |
| **@mkanoor** | Need Kafka Credential type with all new options | Credential type YAML added at `credential_types/kafka-avro-schema-registry.yml` and gist link updated |

---

### Test Results

- **94 unit tests passing** (was 91 before this round of fixes)
- **kafka.py coverage: 93%**
- All pre-commit hooks pass (pylint, mypy, ruff, flake8, black, isort)

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -1,32 +1,53 @@
...
-  - You can override this by providing a valid UUID via the "message_uuid"
-    header or the "message_uuid" field in the message body. Headers take
+  - You can override this by providing a valid UUID via the C(message_uuid)
+    header or the C(message_uuid) field in the message body. Headers take
     precedence over body.
-  - If a provided "message_uuid" is not a valid UUID, the generated UUID is
-    used instead and the original value is preserved under meta.message_id.
+  - If a provided C(message_uuid) is not a valid UUID, the generated UUID is
+    used instead and the original value is preserved under C(meta.message_id).
+  - Messages with a null value (Kafka tombstones) are skipped.
```
ptoscano, line 47: These changes are unrelated to Avro; please move them to a separate PR (as this is already big enough).

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -153,6 +174,77 @@
     description:
       - Timeout in seconds to wait for feedback from the rule engine
         before raising an exception. Only applies when feedback is enabled.
+  message_format:
+    description:
+      - The deserialization format for Kafka message values.
+      - When set to C(json), messages are decoded as UTF-8 text and parsed as JSON (default behavior).
+      - When set to C(avro), messages are deserialized using Apache Avro.
```
ptoscano: The two `C(...)` are option values, so they need to be `V(...)`.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -153,6 +174,77 @@
...
+        Requires the C(fastavro) Python package to be installed.
+      - With Avro, the plugin first attempts to use a local schema file if C(avro_schema_file) is provided.
```
ptoscano: `C(avro_schema_file)` is an option, so `O(...)`.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -153,6 +174,77 @@
...
+    description:
+      - When C(true), reuse the Kafka SSL settings (C(cafile), C(certfile), C(keyfile))
+        for Schema Registry HTTPS connections.
+    type: bool
+    default: true
+notes:
+  - >-
+    A custom EDA credential type for this plugin is available as a GitHub Gist
+    at U(https://gist.github.com/B-Whitt/7b4d08dbc8b78a706b5db4a5ad00e22d). It uses C(extra_vars) and C(file) injectors
+    to securely manage Kafka, SSL, Avro, and Schema Registry parameters.
```
ptoscano, line 265: I guess this needs to be removed.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+        Raises:
+            ImportError: If fastavro is not installed.
+            FileNotFoundError: If schema file does not exist.
+            ValueError: If configuration is invalid.
+
+        """
+        if not HAS_FASTAVRO:
+            msg = (
+                "The 'fastavro' package is required for Avro deserialization. "
+                "Install it with: pip install fastavro"
```
ptoscano: The `fastavro` requirement for Avro is better documented as `notes:` in the `DOCUMENTATION` block:
https://docs.ansible.com/projects/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+        self._logger = logging.getLogger()
+        self._schema: dict[str, Any] | None = None
+
+        # Schema Registry state
+        self._registry_url = schema_registry_url
+        # mypy no-any-return: typed values to match return type
+        self._schema_cache: OrderedDict[int, dict[str, Any]] = OrderedDict()
+        self._http_session: aiohttp.ClientSession | None = None
+        self._ssl_context: ssl.SSLContext | None = None
+        self._request_timeout = aiohttp.ClientTimeout(total=30)
```
ptoscano: This is never used as instance variable, make it a class variable instead (like `_WIRE_FORMAT_MAGIC`, etc).

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+            self._schema_cache.move_to_end(schema_id)
+            return self._schema_cache[schema_id]
+
+        # Fetch from registry
+        url = f"{self._registry_url}/schemas/ids/{schema_id}"
+        headers = await self._get_auth_headers()
+
+        try:
+            if self._http_session is None or self._http_session.closed:
+                self._http_session = aiohttp.ClientSession()
```
ptoscano: `_http_session` is used only in this function to fetch the schema; since this function is called by `_get_schema_from_registry()`, simply inline the creation of this `ClientSession` instance directly in `_get_schema_with_retry()` (like done in `_fetch_oauth_token()`), to avoid managing this instance variable (and the need to `close()` it).

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+        return None
+
+    def _load_schema_file(self, schema_path: str) -> None:
+        """Load and validate a local Avro schema file."""
+        # pathlib over os.path (PTH111/PTH113)
+        resolved = Path(schema_path).expanduser().resolve()
+
+        if not resolved.is_file():
+            msg = f"Avro schema file not found: {resolved}"
+            raise FileNotFoundError(msg)
```
ptoscano: This is a bit redundant, as the `resolved.open()` that follows will fail anyway for non-existing files.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+            ssl_keyfile: Client key file for SSL.
+            ssl_password: Password for encrypted keyfiles.
+
+        Raises:
+            ImportError: If fastavro is not installed.
+            FileNotFoundError: If schema file does not exist.
+            ValueError: If configuration is invalid.
+
+        """
+        if not HAS_FASTAVRO:
```
ptoscano: Rather than raising in this constructor, better move the check directly in `_configure_avro()`.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+            "skipping message: %s",
+            schema_id,
+            self._REGISTRY_MAX_RETRIES,
+            last_exc,
+        )
+        return None
+
+    def _load_schema_file(self, schema_path: str) -> None:
+        """Load and validate a local Avro schema file."""
+        # pathlib over os.path (PTH111/PTH113)
```
ptoscano: Redundant comment...

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+        if self._http_session and not self._http_session.closed:
+            await self._http_session.close()
+
+
+def _configure_avro(
+    args: dict[str, Any],
+    ssl_cafile: str | None = None,
+    ssl_certfile: str | None = None,
+    ssl_keyfile: str | None = None,
+    ssl_password: str | None = None,
```
ptoscano: The parameters for this helper function are inconsistent with what in use so far in the module; in particular, the module arguments are already "parsed" by the helper `_parse_kafka_args()`. So:
- add also the new arguments to the dict returned by `_parse_kafka_args()` (with defaults)
- move some of the basic parameter validation to `_parse_kafka_args()`; for example the check of `message_format`
- pass the actual arguments `_configure_avro()` instead of `args`; follow what `_create_ssl_context()` does

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+) -> tuple[str, AvroDeserializer | None]:
+    """Validate and configure Avro deserialization from plugin args."""
+    logger = logging.getLogger()
```
ptoscano: Rather than creating a logging object here and in AvroDeserializer, make it a module-level variable instead; see what the `webhook` event source does.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+                else:
+                    self._logger.debug(
+                        "Ignoring additional record(s) in Object Container message",
+                    )
+                    break
+        except Exception:  # pylint: disable=broad-exception-caught
+            self._logger.exception(
+                "Failed to deserialize Avro message as Object Container format",
+            )
+            return None
```
ptoscano: Limit this try..except block to only the fastavro usage; pylint is correct here!

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -153,6 +174,77 @@
     description:
       - Timeout in seconds to wait for feedback from the rule engine
         before raising an exception. Only applies when feedback is enabled.
+  message_format:
+    description:
+      - The deserialization format for Kafka message values.
+      - When set to C(json), messages are decoded as UTF-8 text and parsed as JSON (default behavior).
+      - When set to C(avro), messages are deserialized using Apache Avro.
+        Requires the C(fastavro) Python package to be installed.
```
mkanoor: @B-Whitt the fastavro python package might have to be made into an RPM if we are going to backport this to 2.6. In 2.7 we have Konflux to build DE's which does things differently.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -153,6 +174,77 @@
...
+  schema_registry_ssl:
+    description:
+      - When C(true), reuse the Kafka SSL settings (C(cafile), C(certfile), C(keyfile))
+        for Schema Registry HTTPS connections.
+    type: bool
+    default: true
+notes:
+  - >-
+    A custom EDA credential type for this plugin is available as a GitHub Gist
+    at U(https://gist.github.com/B-Whitt/7b4d08dbc8b78a706b5db4a5ad00e22d). It uses C(extra_vars) and C(file) injectors
```
mkanoor: @B-Whitt Shouldn't this point to the plugins/event_source/credential_types/kafka-avro-schema-registry.yaml?

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+            avro_schema_file: Path to a local .avsc schema file.
+            schema_registry_url: Schema Registry URL (http/https).
+            schema_registry_basic_auth: Basic auth credentials (user:password).
+            schema_registry_bearer_token: Static bearer/JWT token.
+            schema_registry_oauth_client_id: OAuth 2.0 client ID.
+            schema_registry_oauth_client_secret: OAuth 2.0 client secret.
+            schema_registry_oauth_token_url: OAuth 2.0 token endpoint URL.
+            schema_registry_oauth_scope: OAuth 2.0 scope.
+            schema_registry_ssl: Reuse Kafka SSL settings for Registry.
+            ssl_cafile: CA file for SSL connections.
```
mkanoor: @B-Whitt If the ssl_cafile/ssl_certfile/ssl_keyfile/ssl_password are for Schema Registry then we should change it prefix all of them with **schema_** to be consistent with other parameters. We might have a similar issue when we get OAuth2 tokens but I am hoping they would be signed by a well known CA in the default OS's trust chain.
We have multiple Servers that we connect to here
* Kafka
* Schema Registry
* OAuth2 Server optionally
And they all have their own certs and keys so we need to keep them separate.
I think the kafka ones are called cafile et al.

requirements.txt
```diff
@@ -3,6 +3,7 @@ pyyaml>=6.0.1
 aiobotocore
 aiohttp
 aiokafka[gssapi]
+fastavro
```
ptoscano, line 6: This effectively makes `fastavro` mandatory, so in that case the logic for handling it as optional package is a bit redundant...

ptoscano (dismissed):
Initial review for the code only, as I haven't got the setup to test it yet.

Regarding the markup markers in documentation (e.g. `C(...)`), I only noted a couple of places; please check the remaining ones, using the following doc as reference:
https://docs.ansible.com/projects/ansible/latest/dev_guide/ansible_markup.html

Thanks!

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -171,13 +263,592 @@
...
+        with resolved.open(encoding="utf-8") as f:
+            raw_schema = json.load(f)
+
+        self._schema = fastavro.parse_schema(raw_schema)
+        self._logger.info("Loaded Avro schema from %s", resolved)
+
+    async def deserialize(self, data: bytes) -> dict[str, Any] | None:
+        """Deserialize Avro binary data.
+
+        Attempts deserialization in priority order:
```
mkanoor, line 746: @B-Whitt Have we been able to test these different wire formats

extensions/eda/plugins/event_source/credential_types/kafka-avro-schema-registry.yml
```diff
@@ -0,0 +1,224 @@
+---
```
mkanoor: @B-Whitt To prevent errors during creating the Credential Type we should create a separate yml file for inputs and injectors when we copy the fields under inputs and paste it in the Credential Type inputs screen it fails with an error message "schema must be in dict format"

extensions/eda/plugins/event_source/credential_types/kafka-avro-schema-registry.yml
```diff
@@ -0,0 +1,224 @@
...
+      help_text: OAuth 2.0 client secret for Schema Registry authentication.
+    - id: schema_registry_oauth_token_url
+      type: string
+      label: Schema Registry OAuth Token URL
+      help_text: OAuth 2.0 token endpoint URL for obtaining access tokens.
+    - id: schema_registry_oauth_scope
+      type: string
+      label: Schema Registry OAuth Scope
+      help_text: OAuth 2.0 scope to request when obtaining access tokens.
+    - id: schema_registry_ssl
```
mkanoor: @B-Whitt We are missing the schema_registry_cafile and related fields here

extensions/eda/plugins/event_source/credential_types/kafka-avro-schema-registry.yml
mkanoor: Image

extensions/eda/plugins/event_source/kafka.py
B-Whitt: done

extensions/eda/plugins/event_source/kafka.py
B-Whitt: updated

extensions/eda/plugins/event_source/kafka.py
B-Whitt: done

extensions/eda/plugins/event_source/kafka.py
B-Whitt: I had to update this as a result of rebasing on top of all the kafka changes with event persistence. Are you requesting that the comments be removed?

extensions/eda/plugins/event_source/kafka.py
B-Whitt: updated

extensions/eda/plugins/event_source/kafka.py
B-Whitt: updated

extensions/eda/plugins/event_source/kafka.py
B-Whitt: added to notes

extensions/eda/plugins/event_source/kafka.py
B-Whitt: updated

extensions/eda/plugins/event_source/credential_types/kafka-avro/injectors.yml
```diff
@@ -0,0 +1,39 @@
...
+  kafka_avro_port: "{{ port }}"
+  kafka_avro_topic: "{{ topic }}"
+  # Optional fields — must have defaults to avoid StrictUndefined
+  kafka_avro_group_id: "{{ group_id | default('', true) }}"
+  kafka_avro_offset: "{{ offset | default('latest', true) }}"
+  security_protocol: "{{ security_protocol | default('PLAINTEXT', true) }}"
+  sasl_mechanism: "{{ sasl_mechanism | default('PLAIN', true) }}"
+  sasl_plain_username: "{{ sasl_plain_username | default('', true) }}"
+  sasl_plain_password: "{{ sasl_plain_password | default('', true) }}"
+  kafka_avro_cafile: "{{ eda.filename.cafile | default('', true) }}"
```
mkanoor: @B-Whitt  None of the eda.filename.xxxx can be in the extra vars section, they only exist in the file section of the injectors and you can access the file name directly. e.g

file:
    template.**example_file**: "{{ example_file_contents }}"


ansible-rulebook will automatically save the contents to a temporary location and create a variable call **eda.filename.example_file** so you dont need for it to be listed in extra_vars since those variables cannot be resolved on the eda-server side.

extensions/eda/plugins/event_source/kafka.py
B-Whitt: Yes. I have not heard that we will be backporting this, but if so I will address the RPM

extensions/eda/plugins/event_source/kafka.py
ptoscano: More than then comments, the changes to those comments.

extensions/eda/plugins/event_source/credential_types/kafka-avro-schema-registry.yml
B-Whitt: files separated and added to extensions/eda/plugins/event_source/credential_types/kafka-avro

extensions/eda/plugins/event_source/kafka.py
B-Whitt: yes. i have tested these 3 use cases.

extensions/eda/plugins/event_source/kafka.py
ptoscano: Please remove the installation notes from here then. Also, see my other comment in `requirements.txt`: is fastavro mandatory or not?

extensions/eda/plugins/event_source/kafka.py
B-Whitt: removed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: addressed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: addressed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: addressed

requirements.txt
B-Whitt: removed redundant code

extensions/eda/plugins/event_source/kafka.py
B-Whitt: removed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: removed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: resolved

extensions/eda/plugins/event_source/kafka.py
B-Whitt: removed

extensions/eda/plugins/event_source/kafka.py
B-Whitt: Done. Reverted the 3 C() formatting changes on UUID lines. Avro-related additions remain.

extensions/eda/plugins/event_source/credential_types/kafka-avro-schema-registry.yml
B-Whitt: added

extensions/eda/plugins/event_source/credential_types/kafka-avro/injectors.yml
B-Whitt: @mkanoor 7 eda.filename.* lines from injectors.yml extra_vars

extensions/eda/rulebooks/demo_avro-local-test.yml
```diff
@@ -0,0 +1,35 @@
+---
+# Parameterized rulebook for EDA Controller UI testing.
+# Used by: Use Case 2a (Local Schema File)
+# Variable names match the kafka-avro credential type injectors.yml.
+# The avro_schema_file path comes from the credential type's file injector,
+# which writes the schema JSON content to a temp file at runtime.
+- name: "Avro + Local Schema File Test"
+  hosts: localhost
+  match_multiple_rules: true
```
mkanoor: @B-Whitt We dont need this here we are not trying to matching multiple rules for the same event. Its an advanced use case and we shouldn't use it in examples.

extensions/eda/rulebooks/demo_avro-ocf-test.yml
```diff
@@ -0,0 +1,33 @@
+---
+# Parameterized rulebook for EDA Controller UI testing.
+# Used by: Use Case 2b (Object Container Format)
+# Variable names match the kafka-avro credential type injectors.yml.
+# No schema file or registry needed — the schema is embedded in each message.
+- name: "Avro Object Container Format Test"
+  hosts: localhost
+  match_multiple_rules: true
```
mkanoor: @B-Whitt We dont need this here we are not trying to matching multiple rules for the same event. Its an advanced use case and we shouldn't use it in examples.

extensions/eda/rulebooks/demo_avro-sr-test.yml
```diff
@@ -0,0 +1,45 @@
+---
+# Parameterized rulebook for EDA Controller UI testing.
+# Used by: Use Cases 1, 3a, 3b, 3c (Schema Registry + all auth variants)
+# All parameters are injected via credential type extra_vars.
+# Variable names match the kafka-avro credential type injectors.yml.
+# Unused auth params are handled by EDA controller (undefined optional fields are safe).
+- name: "Avro + Schema Registry Test"
+  hosts: localhost
+  match_multiple_rules: true
```
mkanoor: @B-Whitt We dont need this here we are not trying to matching multiple rules for the same event. Its an advanced use case and we shouldn't use it in examples.

extensions/eda/rulebooks/kafka-test-schema-registry.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Test Kafka Avro with Schema Registry
+  hosts: localhost
+  match_multiple_rules: true
```
mkanoor: @B-Whitt We dont need this here we are not trying to matching multiple rules for the same event. Its an advanced use case and we shouldn't use it in examples.

extensions/eda/rulebooks/demo_avro-local-test.yml
B-Whitt: removed

extensions/eda/rulebooks/demo_avro-ocf-test.yml
B-Whitt: removed

extensions/eda/rulebooks/demo_avro-sr-test.yml
B-Whitt: removed

extensions/eda/rulebooks/kafka-test-schema-registry.yml
B-Whitt: removed

mkanoor (approved):

extensions/eda/plugins/event_source/credential_types/kafka-avro/injectors.yml
```diff
@@ -0,0 +1,32 @@
...
+  sasl_plain_username: "{{ sasl_plain_username }}"
+  schema_registry_basic_auth: "{{ schema_registry_basic_auth }}"
+  schema_registry_bearer_token: "{{ schema_registry_bearer_token }}"
+  schema_registry_oauth_client_id: "{{ schema_registry_oauth_client_id }}"
+  schema_registry_oauth_client_secret: "{{ schema_registry_oauth_client_secret }}"
+  schema_registry_oauth_scope: "{{ schema_registry_oauth_scope }}"
+  schema_registry_oauth_token_url: "{{ schema_registry_oauth_token_url }}"
+  schema_registry_ssl: "{{ schema_registry_ssl }}"
+  schema_registry_url: "{{ schema_registry_url }}"
+  schema_ssl_password: "{{ schema_registry_password }}"
```
wfealdel, line 30: **Naming asymmetry:** The input field in `input.yml` is `schema_registry_password` but the extra_var here is `schema_ssl_password`. Every other field in this file uses matching names on both sides. Consider renaming the input field in `input.yml` to `schema_ssl_password` to align the chain end-to-end.

```suggestion
  schema_ssl_password: "{{ schema_ssl_password }}"
```

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -398,12 +1168,45 @@ async def main(
...
+
+    if message_format == "avro" and deserializer is not None:
+        data = await deserializer.deserialize(msg_value)
+        if data is None:
+            logger.warning("Avro deserialization returned None, skipping message")
+        return data
+
+    try:
+        value = msg_value.decode(encoding)
+        return cast("dict[str, Any]", json.loads(value))
```
wfealdel, line 1201: **Type safety:** `json.loads()` can return `list`, `int`, `bool`, etc. — not just `dict`. The `cast` hides this from the type checker, and the return type annotation `dict[str, Any] | str | None` does not account for lists.

Downstream code already handles non-dict bodies correctly (`isinstance(data, dict)` guard in `_process_event_uuid`), so removing the cast and widening the return type of `_decode_message_body` to `Any` would be the minimal fix.

```suggestion
        return json.loads(value)
```

wfealdel (commented):
Well-engineered feature with thorough test coverage (94 tests). Two minor inline comments on naming consistency and type safety — see threads above. No blocking issues found.

wfealdel (approved):

extensions/eda/plugins/event_source/kafka.py
B-Whitt: Non blocking comments will be addressed in a later PR.

extensions/eda/plugins/event_source/credential_types/kafka-avro/injectors.yml
B-Whitt: Non blocking comments will be addressed in a later PR.

PR #532, opened 2026-04-08
fix(ci): use explicit path in release-drafter _extends directive

B-Whitt:
release-drafter v7 no longer infers the config file path from a bare repo reference, breaking the ack CI check. Use the explicit path format matching the fix in ansible/ansible-creator#577.

Assisted by: Claude

## Summary by CodeRabbit

* **Chores**
  * Updated release management configuration to reference a different upstream release-drafter template path, changing the base template used for automated release drafts. This is an internal tooling update and does not modify public APIs or exported entities.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,3 +1,3 @@
 ---
 # see https://github.com/ansible/team-devtools
-_extends: ansible/team-devtools
+_extends: ansible/team-devtools:/.github/release-drafter.yml


ptoscano (approved):

andresberejnoi (approved):

ptoscano:
Ah, I accidentally rebased it through github, which of course invalidated the signature of the commit; please sign this commit again and rebase it, thanks! (and sorry)

PR #542, opened 2026-05-15
fix(tests) [AAP-75500] Add retry logic to deleting a project. Assisted-by: Claude

jcraiglo1:
https://redhat.atlassian.net/browse/AAP-75500

Lately Tox staging tests have been flaky because of a change to how EDA Server handles project deletes right after creating the project. Because of this change the Delete Project test [here](https://github.com/ansible/event-driven-ansible/blob/main/tests/integration/targets/project/tasks/create.yml#L64-L73) is flaky. 

This changes a test that deletes a project to have retries. This should eliminate any race conditions we might be facing. 

NOTE: Previously, this issue isn't only on this test. This PR https://github.com/ansible/event-driven-ansible/pull/531 was created to address the same issue in another test. I am using that same strategy here.  With that being said please share your thoughts and I am more than happy to rescind the PR and instead update the tests. Thank you! 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of project deletion by implementing automatic retry logic. The module now attempts to delete projects up to 5 times with recovery attempts between failures, providing better handling of transient errors during the deletion process.

### DIFF
--- a/plugins/modules/project.py
+++ b/plugins/modules/project.py
@@ -314,26 +314,11 @@ def main() -> None:
         )
 
     if state == "absent":
-        # Retry delete when the server rejects it due to an active job
-        # (e.g. a sync that just finished but whose background work is
-        # still being cleaned up).
-        max_retries = 5
-        retry_delay = 3
-        for attempt in range(max_retries):
-            try:
-                ret = controller.delete_if_needed(project, endpoint=project_endpoint)
-                break
-            except EDAError as eda_err:
-                if attempt < max_retries - 1:
-                    time.sleep(retry_delay)
-                    try:
-                        project = controller.get_exactly_one(
-                            project_endpoint, name=project_name
-                        )
-                    except EDAError:
-                        pass
-                else:
-                    module.fail_json(msg=str(eda_err))
+        # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
+        try:
+            ret = controller.delete_if_needed(project, endpoint=project_endpoint)
+        except EDAError as eda_err:
+            module.fail_json(msg=str(eda_err))
         module.exit_json(**ret)
 
     project_params: dict[str, Any] = {}
--- a/plugins/modules/project.py
+++ b/plugins/modules/project.py
@@ -314,11 +314,26 @@ def main() -> None:
         )
 
     if state == "absent":
-        # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
-        try:
-            ret = controller.delete_if_needed(project, endpoint=project_endpoint)
-        except EDAError as eda_err:
-            module.fail_json(msg=str(eda_err))
+        # Retry delete when the server rejects it due to an active job
+        # (e.g. a sync that just finished but whose background work is
+        # still being cleaned up).
+        max_retries = 5
+        retry_delay = 3
+        for attempt in range(max_retries):
+            try:
+                ret = controller.delete_if_needed(project, endpoint=project_endpoint)
+                break
+            except EDAError as eda_err:
+                if attempt < max_retries - 1:
+                    time.sleep(retry_delay)
+                    try:
+                        project = controller.get_exactly_one(
+                            project_endpoint, name=project_name
+                        )
+                    except EDAError:
+                        pass
+                else:
+                    module.fail_json(msg=str(eda_err))
         module.exit_json(**ret)
 
     project_params: dict[str, Any] = {}
--- a/tests/integration/targets/project/tasks/create.yml
+++ b/tests/integration/targets/project/tasks/create.yml
@@ -61,11 +61,23 @@
     that:
       - r.changed
 
+- name: Wait for project sync to complete before delete
+  ansible.eda.project_info:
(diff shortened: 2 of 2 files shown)


jcraiglo1:
Slack thread talking about the issue: https://redhat-internal.slack.com/archives/C053CJ4JFP1/p1777983100694219Connect your Slack account 

Example of the failing test: [tox · ansible/event-driven-ansible@dd47e63](https://github.com/ansible/event-driven-ansible/actions/runs/25794744856/job/75769209347#step:9:1899)

plugins/modules/project.py
```diff
@@ -314,11 +314,26 @@ def main() -> None:
...
-        # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
-        try:
-            ret = controller.delete_if_needed(project, endpoint=project_endpoint)
-        except EDAError as eda_err:
-            module.fail_json(msg=str(eda_err))
+        # Retry delete when the server rejects it due to an active job
+        # (e.g. a sync that just finished but whose background work is
+        # still being cleaned up).
+        max_retries = 5
+        retry_delay = 3
```
kaiokmo: Does it make sense to add this logic here in the module versus adding retry/delay to the ansible task parameters itself? I mean, the error when deleting a project that hasn't finished being imported is something that the EDA API returns because it's by design, right?

That way, I think we shouldn't change the module to try to work around the API design, but instead, as users of the collection, if the API returned that the project cannot be deleted because it's still being imported, we should address it by retrying in the calling task itself.

WDYT?

jcraiglo1: The error from EDA-server is by design, yes. By adding the retries into EDA we make it more robust for cheap. This also allows us to keep our tests cleaner. Instead of adding to them because of a side effect in EDA-Server.

With that being said, I can see your argument of keeping EDA's design decisions agnostic of EDA-servers quirks. But if we're already basing our tests on EDA-Server, why not improve the function of EDA itself? 

I mentioned this in the description but I really do think either way of thinking is fine and I'm happy to do it however the team agrees

ptoscano: > as users of the collection, if the API returned that the project cannot be deleted because it's still being imported, we should address it by retrying in the calling task itself.

I don't disagree with it. I only note that right now in the failed case there is only the error message, so users have no choice than retrying on failure no matter the failure. I imagine suggesting to retry is OK when you know the issue is a transient one such as the syncing still in progress (as it will end), compared to other ones (e.g. auth issue, missing project, etc). Maybe we should return the last HTTP status code when failing too?

> By adding the retries into EDA we make it more robust for cheap.

Not sure what would EDA need to do differently here; the flow is this:

1. in EDA, something starts the sync of a project
2. while the project is still sync'ing (e.g. it takes time to pull the Git repository), a client issues a DELETE request for that project
3. EDA returns 409 as deleting the project while a sync is ongoing is not possible

Possible things that I see changing are:
- actually abort the project sync during a DELETE
  - this is tricky to do, and I think the current behaviour is a result of avoiding EDA not handling this well in the past
- make the DELETE synchronous, actually waiting for the project sync to complete
  - I think this likely breaks any kind of REST API design, not to mention the sync could take a long time (even getting into HTTP timeouts)

So IMHO the current situation is the best we have, all things considered. That said, if we want this change, IMHO it ought to be done for all the other modules too, otherwise there is inconsistency in the behaviour of the modules.

AlexSCorey: Ahh the age old problem of waiting for projects to sync so we can run tests.  This is problem I've experienced on the controller team, and the UI team.  The best solution we came up with was reuse projects as much as possible via some "default" project that was already synced.  This project should be used when testing things that are not really project related, but require a project.  For project tests we could use some naming convention to indicate a project is used for only a single test and after all the tests are done we could clean it all up.  Both of those implementation require a rather robust rewrite if not already implemented.  For this use case I sort of agree that this might be the best we can do.  Changes to help improve on this should probably be done on the API side.

jcraiglo1: I have updated the Pr to instead add retries to the tests

AlexSCorey (approved):
Looks like all my concerns have been addressed.  LGTM

tests/integration/targets/project/tasks/create.yml
```diff
@@ -61,11 +61,23 @@
     that:
       - r.changed
 
+- name: Wait for project sync to complete before delete
+  ansible.eda.project_info:
+    name: "{{ project_name }}"
+  register: project_info
+  until: project_info.projects[0].import_state == "completed"
+  failed_when: project_info.projects[0].import_state == "failed"
```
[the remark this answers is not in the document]
ptoscano: @coderabbitai If `project_info.projects` is empty then there is a different issue which needs to be debugged and fixed, so adding checks would only hide it. Better fail in that case so we notice it is actually an issue.

ptoscano (approved):

ptoscano:
@jcraiglo1 would you please rebase this squashing the commits into one, and amend the commit message and the PR description saying the actual changes being done?
Thanks!

PR #545, opened 2026-05-21
fix: add UUID validation to kafka source plugin

ttuffin:
Currently we perform no UUID validation on the kafka event source plugin, which is causing validation failures in eda-server. This fix implements UUID generation in the event message_uuid is not a valid UUID.

Assisted-by: Claude

## Summary by CodeRabbit

* **New Features**
  * Deterministic UUID assignment for Kafka events (UUID5 from topic:partition:offset). A valid UUID in headers overrides body; invalid header UUIDs trigger a one-time warning, are preserved as the original identifier, and the generated UUID is used.

* **Documentation**
  * Clarified UUID assignment rules, precedence, and invalid-UUID fallback behavior.

* **Tests**
  * Added/updated tests for deterministic UUIDs, header/body precedence, and invalid-UUID fallback.

### DIFF
--- a/extensions/eda/plugins/event_source/kafka.py
+++ b/extensions/eda/plugins/event_source/kafka.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
+import uuid
 from datetime import datetime, timezone
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
 from typing import TYPE_CHECKING, Any
@@ -18,10 +19,14 @@ DOCUMENTATION = r"""
 short_description: Receive events via a kafka topic.
 description:
   - An ansible-rulebook event source plugin for receiving events via a kafka topic.
-  - To make each message unique you can either add message_uuid to the header or
-  - add message_uuid to the message body
-  - if both of these are missing we will use the {topic}:{partition}.{offset} to
-  - make the unique message uuid
+  - Each event is assigned a valid RFC 4122 UUID under meta.uuid.
+  - By default, a deterministic UUID5 is generated from the Kafka message
+    coordinates (topic, partition, offset).
+  - You can override this by providing a valid UUID via the "message_uuid"
+    header or the "message_uuid" field in the message body. Headers take
+    precedence over body.
+  - If a provided "message_uuid" is not a valid UUID, the generated UUID is
+    used instead and the original value is preserved under meta.message_id.
 options:
   host:
     description:
@@ -170,6 +175,7 @@ EXAMPLES = r"""
 
 MESSAGE_UUID_KEY = "message_uuid"
 QUEUE_TIMEOUT = 120
+_uuid_warning = {"emitted": False}
 
 
 def _validate_and_normalize_topics(args: dict[str, Any]) -> list[str] | None:
@@ -330,6 +336,15 @@ def get_message_timestamp(msg: ConsumerRecord) -> str:
     )
 
 
+def is_valid_uuid(value: str) -> bool:
+    """Check if a string is a valid RFC 4122 UUID."""
+    try:
+        uuid.UUID(value)
+        return True  # noqa: TRY300
+    except (ValueError, AttributeError):
+        return False
+
+
 async def main(
     queue: asyncio.Queue[Any],
     args: dict[str, Any],
@@ -389,6 +404,52 @@ async def main(
         await kafka_consumer.stop()
 
 
+def _process_event_uuid(
+    msg: ConsumerRecord,
+    headers: dict[str, str],
+    data: dict[str, Any] | str | None,
+    event: dict[str, Any],
+) -> None:
+    """Process the event UUID from headers, body, or Kafka coordinates.
+
+    Priority: header message_uuid > body message_uuid > generated UUID5.
+    If a provided message_uuid is not a valid UUID, the generated UUID is
+    used and the original value is preserved under event.meta.message_id.
+    """
+    logger = logging.getLogger()
+
+    event_uuid = str(
+        uuid.uuid5(
+            uuid.NAMESPACE_OID,
+            f"{msg.topic}:{msg.partition}:{msg.offset}",
+        ),
+    )
+
+    message_id = None
+    if MESSAGE_UUID_KEY in headers:
+        message_id = headers[MESSAGE_UUID_KEY]
+    elif isinstance(data, dict) and MESSAGE_UUID_KEY in data:
+        message_id = data[MESSAGE_UUID_KEY]
+
+    if message_id:
+        if is_valid_uuid(message_id):
+            event_uuid = message_id
+        else:
+            # If not a valid UUID, we must warn the user and use
+            # the generated UUID instead. The original UUID is copied to
+            # event.meta.message_id for tracking purposes.
+            event["meta"]["message_id"] = message_id
+            if not _uuid_warning["emitted"]:
+                _uuid_warning["emitted"] = True
+                logger.warning(
+                    "Provided message_uuid is not a valid UUID,"
+                    " using generated UUID. The original value has been"
+                    " stored under event.meta.message_id for tracking.",
+                )
+
+    event["meta"]["uuid"] = event_uuid
+
+
 async def receive_msg(
     queue: asyncio.Queue[Any],
     kafka_consumer: AIOKafkaConsumer,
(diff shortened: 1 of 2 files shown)


extensions/eda/plugins/event_source/kafka.py
```diff
@@ -428,17 +433,41 @@ async def receive_msg(
...
+        event_uuid = str(
+            uuid.uuid5(
+                uuid.NAMESPACE_OID,
+                f"{msg.topic}:{msg.partition}:{msg.offset}",
+            )
+        )
+
+        message_uuid = None
+        if MESSAGE_UUID_KEY in headers:
+            message_uuid = headers[MESSAGE_UUID_KEY]
```
mkanoor: @ttuffin I think this should just be called message_id since we dont know its format. It could be uuid or something else.

ptoscano: The actual field set for the event is named `message_id` already, so it's only the Python variable misleading.

ptoscano:
In addition to what Madhu wrote, please update the documentation -- in particular `description` in `DOCUMENTATION`.

extensions/eda/plugins/event_source/kafka.py
ttuffin: Changed `message_uuid` to `message_id`

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -402,24 +461,20 @@ async def receive_msg(
...
-            headers: dict[str, str] = {
-                header[0]: header[1].decode(encoding) for header in msg.headers
-            }
+            headers = {header[0]: header[1].decode(encoding) for header in msg.headers}
             event["meta"]["headers"] = headers
-            if MESSAGE_UUID_KEY in headers:
-                event["meta"]["uuid"] = headers[MESSAGE_UUID_KEY]
-
         except UnicodeError:
             logger.exception("Unicode error while decoding headers")
```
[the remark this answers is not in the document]
ptoscano: @coderabbitai: are you sure that headers of consumer records can have null values? I haven't find references to that possibility anywhere in official docs for both kafma producers and consumers.

Also, the code before this change had the same behaviour, so technically there is no regression here.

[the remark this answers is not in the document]
ptoscano: @coderabbitai: Yes, please. Thanks anyway.

ptoscano:
The changes generally LGTM.

There are a couple of minor points to mention that I don't consider blocking:
- I find the new `_resolve_event_uuid()` helper inconsistent: it returns something that its caller sets in `event`, while the function itself changes `event` too; the option I see requiring less changes from what done so far is to rename the function to e.g. `_handle_event_uuid` and also set `event["meta"]["uuid"]` there too at the end of it
- in the documentation block, apply the ansible-doc markup to the `message_uuid` & `meta.message_id` references, e.g.: `C(message_uuid)`; reference: https://docs.ansible.com/projects/ansible/latest/dev_guide/ansible_markup.html

Also, should `meta.message_id` be always filled, no matter what was used as `meta.uuid`?

ttuffin:
@ptoscano I have updated and renamed the helper function to reflect its job of processing the event UUID. Regarding the docs, that was how I originally implemented it but after checking the docs in the existing source plugins I see we do not use `C(...)` in any of them, so I have opted to keep it consistent. Better to update all the plugin docs at the same time in a separate PR.

ptoscano (dismissed):

ttuffin:
Here is the results of tests run with this implementation:
**Valid message_uuid**
```
'id': 1,
'message': 'Hello from batch '
'producer - message 1',
'message_uuid': '9bbc6177-c14d-451b-86d8-dfba6929c52a',
'timestamp': '2026-05-26T10:36:06.816178'},
'meta': {'headers': {},
'produced_at': '2026-05-26T08:36:06.816000Z',
'received_at': '2026-05-26T08:36:06.847204Z',
'source': {'name': 'kafka_test',
'type': 'ansible.eda.kafka'},
'uuid': '9bbc6177-c14d-451b-86d8-dfba6929c52a'}},
```

**Invalid message_uuid**
```
'id': 7,
'message': 'Hello from batch '
'producer - message 7',
'message_uuid': 'atf-happy-aa168cd6b3a5073e7987272688f9ae96:0:4',
'timestamp': '2026-05-26T10:26:22.900660'},
'meta': {'headers': {},
'message_id': 'atf-happy-aa168cd6b3a5073e7987272688f9ae96:0:4',
'produced_at': '2026-05-26T08:26:22.900000Z',
'received_at': '2026-05-26T08:26:22.987485Z',
'source': {'name': 'kafka_test',
'type': 'ansible.eda.kafka'},
'uuid': '892575e8-0874-57ea-8232-e8120d3ca75b'}},
```

As you can see from the output of the latter, we end up with duplicate values for `message_uuid` and `message_id`. Since it looks like the original value remains preserved in `message_uuid`, I am thinking there is no need to copy it to the new `message_id` key. But perhaps there is a corner case I am not considering?

The warning message shows up as expected, but it is pretty spammy since it prints for every single message with an invalid UUID. If we want to keep it that way I think we should change it to debug level.

`root - WARNING - Provided message_uuid 'atf-happy-aa168cd6b3a5073e7987272688f9ae96:0:4' is not a valid UUID, using generated UUID. The original value has been stored under event.meta.message_id for tracking.`

ptoscano:
> As you can see from the output of the latter, we end up with duplicate values for `message_uuid` and `message_id`. Since it looks like the original value remains preserved in `message_uuid`, I am thinking there is no need to copy it to the new `message_id` key. But perhaps there is a corner case I am not considering?

Hm the other case added so far is that the uuid is now also searched in the data, and not only in the headers. While it could be useful to still have the uuid in `meta`, if you put the uuid in the data then you know about it, so you can look it up yourself in your rulebook conditions? I guess that makes the new meta.message_id` kinda redundant, yes.

> The warning message shows up as expected, but it is pretty spammy since it prints for every single message with an invalid UUID. If we want to keep it that way I think we should change it to debug level.

What about logging only the first message as warning, and any other message as debug? In this case, adding to the current message something about that any following message like this will get the same handling of the uuid.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -389,6 +403,51 @@ async def main(
...
+        ),
+    )
+
+    message_id = None
+    if MESSAGE_UUID_KEY in headers:
+        message_id = headers[MESSAGE_UUID_KEY]
+    elif isinstance(data, dict) and MESSAGE_UUID_KEY in data:
+        message_id = data[MESSAGE_UUID_KEY]
+
+    if message_id:
```
mkanoor, line 434: @ttuffin Should we unconditionally set the event['meta']['message_id'] if the customer has provided one, since customers can use that for closing their loop which we are not privy to.
The event['meta']['uuid'] is for us and drools to prevent duplicate events from getting processed. They can both contain the same value if they use uuid format but if they use some other format then the 2 ways will be different.
We might not be able to force every customer to change it to UUID format but we will preserve their unique id in the message_id field.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -389,6 +403,51 @@ async def main(
...
+        if is_valid_uuid(message_id):
+            event_uuid = message_id
+        else:
+            # If not a valid UUID, we must warn the user and use
+            # the generated UUID instead. The original UUID is copied to
+            # event.meta.message_id for tracking purposes.
+            event["meta"]["message_id"] = message_id
+            logger.warning(
```
mkanoor: @ttuffin Like @ptoscano was mentioning we could do this just once with a static variable so the user knows that they can't use event['meta']['uuid'] to close their loop they have to be dependent on event['meta']['message_id']

extensions/eda/plugins/event_source/kafka.py
ttuffin: @mkanoor `event['meta']['message_id']` is only set if `MESSAGE_UUID_KEY` is defined in the headers or body. If `MESSAGE_UUID_KEY` is undefined, what would we set `event['meta']['message_id']` to?

extensions/eda/plugins/event_source/kafka.py
ttuffin: @mkanoor @ptoscano I've pushed changes so that the warning message is displayed only once. Adding additional messages via debug was still quite spammy, even for debug mode.

extensions/eda/plugins/event_source/kafka.py
```diff
@@ -389,6 +404,52 @@ async def main(
...
+        message_id = data[MESSAGE_UUID_KEY]
+
+    if message_id:
+        if is_valid_uuid(message_id):
+            event_uuid = message_id
+        else:
+            # If not a valid UUID, we must warn the user and use
+            # the generated UUID instead. The original UUID is copied to
+            # event.meta.message_id for tracking purposes.
+            event["meta"]["message_id"] = message_id
```
mkanoor, line 441: @ttuffin looks good I missed this line

ptoscano (approved):

mkanoor (approved):

PR #550, opened 2026-06-12
Users can update a rulebook activation extra_vars and description now.

AlexSCorey:
This resolves [AAP-72266](https://issues.redhat.com/browse/AAP-72266). The issue was that if the state of the rulebook activation did not need to change, meaning it was already editable, we would exit early. Now a user can update a rulebook activation via the collection properly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rulebook activation enable/disable handling so the module only marks a change when a real state transition occurs.
  * When reusing an existing activation, the enabled state is now preserved unless `state` is explicitly set to `enabled` or `disabled`; other supplied fields still override.

* **Tests**
  * Expanded unit and integration coverage for create/update parameter construction, credential/token/k8s re-resolution behavior, source mapping processing, and enable/disable transition expectations (including `changed` outcomes and name propagation).

### DIFF
--- a/plugins/modules/rulebook_activation.py
+++ b/plugins/modules/rulebook_activation.py
@@ -275,7 +275,7 @@ id:
 """
 
 import traceback
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 try:
     import yaml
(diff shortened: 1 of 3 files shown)


tests/integration/targets/activation/tasks/main.yml
```diff
@@ -460,18 +460,19 @@
...
         name: "{{ activation_name_modified }}"
         state: enabled
       register: _result
+      ignore_errors: true
 
     - name: Check rulebook activation not updated
       assert:
         that:
-          - not _result.changed
+          - _result.changed or _result.failed
```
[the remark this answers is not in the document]
AlexSCorey: This assertion was written this way due to flake in the CI.  Please ignore.

plugins/modules/rulebook_activation.py
```diff
@@ -402,92 +402,141 @@ def process_event_streams(
...
+            # Process event streams and source mappings
+            activation_params["source_mappings"] = yaml.dump(
+                process_event_streams(
+                    # ignore type error, if rulebook_id is None, it will fail earlier
+                    rulebook_id=rulebook_id,  # type: ignore[arg-type]
+                    controller=controller,
+                    module=module,
+                )
+            ).rstrip("\n")
+    else:
```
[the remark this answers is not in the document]
jcraiglo1: @AlexSCorey

plugins/modules/rulebook_activation.py
```diff
@@ -656,39 +722,41 @@ def endpoint_state(activation: Dict[str, Any], state: str) -> str:
...
+                state_changed = True
+                activation["is_enabled"] = state == "enabled"
             except EDAError as e:
                 module.fail_json(
                     msg=f"Failed to enable/disable rulebook activation: {e}"
                 )
-
-        module.exit_json(changed=False)
-
+        else:
```
B-Whitt: @AlexSCorey [Functionality] [plugins/modules/rulebook_activation.py:738-740] Early exit for state=enabled when already enabled still blocks parameter updates — the same class of bug this PR fixes for state=disabled.

When state=enabled and the activation is already enabled, the code exits with module.exit_json(changed=False) before reaching create_or_update_if_needed. Changes to extra_vars,
  description, etc. are silently dropped. Meanwhile state=disabled when already disabled correctly falls through and applies updates.

  - Evidence: else: if state == "enabled": module.exit_json(changed=False)
  - Is this finding intentional? Should state=enabled on an already-enabled activation also apply param updates (symmetric with state=disabled), or is the early exit the desired contract?

AlexSCorey: @B-Whitt Good catch.  Thank you for find this.  I've updated

plugins/modules/rulebook_activation.py
```diff
@@ -402,92 +402,156 @@ def process_event_streams(
...
+        # Re-resolve awx_token_name if user provided it
+        if module.params.get("awx_token_name"):
+            awx_token_id = lookup_resource_id(
+                module,
+                controller,
+                "/users/me/awx-tokens/",
+                module.params["awx_token_name"],
+            )
+            if awx_token_id is not None:
+                activation_params["awx_token_id"] = awx_token_id
```
ptoscano: This is practically the same as the "old code" (for a new activation), so I think it can be moved in the common code part of `create_params()`.

plugins/modules/rulebook_activation.py
```diff
@@ -402,92 +402,156 @@ def process_event_streams(
...
+
+        # Re-resolve k8s_service_name if user provided it
+        if not is_aap_24 and module.params.get("k8s_service_name"):
+            activation_params["k8s_service_name"] = module.params["k8s_service_name"]
```
ptoscano: Ditto (same/common code).

plugins/modules/rulebook_activation.py
```diff
@@ -402,92 +402,156 @@ def process_event_streams(
...
+        # Re-resolve credentials if user provided new ones
+        if not is_aap_24 and module.params.get("eda_credentials"):
+            eda_credential_ids = []
+            for item in module.params["eda_credentials"]:
+                cred_id = lookup_resource_id(
+                    module, controller, "eda-credentials", item
+                )
+                if cred_id is not None:
+                    eda_credential_ids.append(cred_id)
+            activation_params["eda_credentials"] = eda_credential_ids
```
ptoscano: Ditto (same/common code).

plugins/modules/rulebook_activation.py
```diff
@@ -402,92 +402,156 @@ def process_event_streams(
...
+            rulebook_id = activation_params.get("rulebook_id")
+            # Process event streams and source mappings
+            activation_params["source_mappings"] = yaml.dump(
+                process_event_streams(
+                    # ignore type error, if rulebook_id is None, it will fail earlier
+                    rulebook_id=rulebook_id,  # type: ignore[arg-type]
+                    controller=controller,
+                    module=module,
+                )
+            ).rstrip("\n")
```
ptoscano: Ditto (same/common code) -- only
```python
            rulebook_id = activation_params.get("rulebook_id")
```
needs to stay in the `if existing:` block.

plugins/modules/rulebook_activation.py
```diff
@@ -656,39 +722,38 @@ def endpoint_state(activation: Dict[str, Any], state: str) -> str:
         ]
         activation["eda_credentials"] = credential_ids
 
+    state_changed = False
     if state and state in ("enabled", "disabled") and activation:
         change_state = endpoint_state(activation=activation, state=state)
-
```
ptoscano, line 661: extra change

plugins/modules/rulebook_activation.py
```diff
@@ -656,39 +722,38 @@ def endpoint_state(activation: Dict[str, Any], state: str) -> str:
...
         if change_state:
             try:
-                # Call disable or enable endpoint
```
ptoscano, line 664: This comment is still valid, isn't it?

plugins/modules/rulebook_activation.py
```diff
@@ -499,10 +563,14 @@ def create_params(
     if module.params.get("restart_policy"):
         activation_params["restart_policy"] = module.params["restart_policy"]
 
-    if module.params.get("state") == "disabled":
-        activation_params["is_enabled"] = False
+    if existing:
+        if module.params.get("state") in ("disabled", "enabled"):
+            activation_params["is_enabled"] = module.params.get("state") != "disabled"
```
ptoscano, line 534: As I see it, this enables automatically an existing activation that was disabled. This can be seen later in the changed task `Assert that rulebook activation is updated and no other activation was created` in `tests/integration/targets/activation/tasks/main.yml`; the task `Enable rulebook activation again` was also changed to ignore the error while trying to enable an activation that was supposed to be present but not enabled.

I don't think this behaviour change is wanted, so IMHO editing an existing activation ought to not change its state.

plugins/modules/rulebook_activation.py
```diff
@@ -579,8 +647,6 @@ def main() -> None:
         restart_on_project_update=dict(type="bool", default=False),
     )
 
-    # Define the state the activation is transitioning to, and uses
-    # this in the call to the according post endpoint.
```
ptoscano, line 583: Still valid comment, isn't it?

ptoscano (changes requested):
The changes are a good starting point, however:
- the code can be simplified a bit more
- the behaviour change IMHO needs to be reverted (see more explanation in comment)

AlexSCorey:
Realized the ticket isn't really a bug and there is an existing work around.  So, closing this pr.

PR #554, opened 2026-06-23
feat(rbac): add role_team_assignment module for AAP 2.5

B-Whitt:
## Summary

- Adds `ansible.eda.role_team_assignment` module for creating and removing
  RBAC team role assignments against eda-server's `/api/eda/v1/role_team_assignments/` endpoint
- Supports name-based or numeric ID resolution for `role_definition` and `team`
- Supports `team_ansible_id` and `object_ansible_id` as UUID alternatives
- States: `present` (idempotent create), `absent` (idempotent delete), `exists` (read-only check)
- Follows the `credential_input_source` composite-key pattern (no `name` field)
- This module is intended for use with AAP 2.5 only. For AAP 2.6+, use ansible.platform.role_team_assignment.

## Context

Jira: [AAP-58542](https://redhat.atlassian.net/browse/AAP-58542)

The `ansible.platform` collection has `role_team_assignment` for AAP 2.6/2.7 but it
was never added to stable-2.5. Backporting is not feasible — the 2.5 branch lacks the
action plugin architecture used by 2.7. This module provides equivalent functionality
in `ansible.eda`, talking directly to eda-server.

Prerequisite server-side change (merged): [aap-eda-controller#220](https://github.com/ansible-automation-platform/aap-eda-controller/pull/220)

## Changes

- `plugins/modules/role_team_assignment.py` — new module
- `meta/runtime.yml` — added to `eda` action group
- `tests/unit/plugins/modules/test_role_team_assignment.py` — 22 unit tests (94% coverage)
- `tests/integration/targets/role_team_assignment/tasks/main.yml` — integration test

## Test plan

- [x] All pre-commit hooks pass (ruff, black, mypy, flake8, pylint, isort)
- [x] Unit tests: 22/22 passed via `tox -e unit`
- [x] Integration test: full lifecycle verified against live AAP 2.5 instance
  - Create assignment (changed=true)
  - Idempotency (changed=false)
  - state=exists returns details
  - Delete (changed=true)
  - Delete idempotency (changed=false)
  - state=exists returns "not found"
  - Non-existent role definition fails with descriptive error

Assisted-By: Claude Opus 4.6

## Release Notes

* **New Features**
  * Added a new `role_team_assignment` module to create, delete, and verify role-to-team assignments in EDA Controller with idempotent behavior and check mode support.
  * Enabled the `role_team_assignment` EDA action in the action group configuration.

* **Tests**
  * Added unit tests covering assignment matching, ID resolution, request payloads/filters, and error handling.
  * Added an integration test playbook validating `present`, `absent`, and `exists` behavior, idempotency, and cleanup.

* **Chores**
  * Updated the runtime action group configuration and included the new module in the manifest.

### DIFF
--- a/plugins/modules/role_team_assignment.py
+++ b/plugins/modules/role_team_assignment.py
@@ -1,0 +1,366 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: Contributors to the Ansible project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: role_team_assignment
+author:
+    - Brandon Whittington (@B-Whitt)
+short_description: Assign a role to a team in EDA Controller
+version_added: "2.13.0"
+description:
+    - This module allows you to create or remove role-based access assignments
+      for teams in EDA Controller.
+    - Assignments grant a team permissions defined by a role definition,
+      optionally scoped to a specific resource (organization, project, etc.).
+    - Assignments are immutable after creation. Delete and re-create to change them.
+    - This module is for AAP 2.5 only. For AAP 2.6+, use
+      ansible.platform.role_team_assignment.
+options:
+    role_definition:
+        description:
+            - The name or numeric ID of the role definition that defines
+              the permissions to grant.
+        required: true
+        type: str
+    team:
+        description:
+            - The name or numeric ID of the team receiving the role.
+            - Mutually exclusive with I(team_ansible_id).
+        required: false
+        type: str
+    team_ansible_id:
+        description:
+            - The Ansible resource UUID of the team.
+            - Alternative to I(team).
+        required: false
+        type: str
+    object_id:
+        description:
+            - The numeric primary key of the resource to scope the assignment to.
+            - Omit for assignments that do not require a specific resource
+              (this depends on the role definition).
+            - Mutually exclusive with I(object_ansible_id).
+        required: false
+        type: int
+    object_ansible_id:
+        description:
+            - The Ansible resource UUID of the resource to scope the assignment to.
+            - Alternative to I(object_id).
+        required: false
+        type: str
+    state:
+        description:
+            - C(present) creates the assignment if it does not exist (idempotent).
+            - C(absent) removes the assignment if it exists (idempotent).
+            - C(exists) checks whether the assignment is present and returns
+              its details. Never modifies anything.
+        choices: ["present", "absent", "exists"]
+        default: "present"
+        type: str
+extends_documentation_fragment: ansible.eda.eda_controller.auths
+notes:
+    - This module is for AAP 2.5 only. For AAP 2.6+, use
+      ansible.platform.role_team_assignment.
+    - Assignments are immutable. They cannot be updated, only created or deleted.
+    - Global roles (content_type=None) cannot be assigned to teams.
+    - Organization Member role cannot be assigned to teams.
+    - The role definition determines which resource type the assignment applies to.
+      The API validates that the object matches the role's content_type.
+"""
+
+
+EXAMPLES = r"""
+- name: Assign Organization Auditor to a team for the Default org
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team: "network-ops"
+    object_id: 1
+    state: present
+
+- name: Assign a role using object_ansible_id for resource scoping
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team: "network-ops"
+    object_ansible_id: "550e8400-e29b-41d4-a716-446655440000"
+    state: present
+
+- name: Remove a team role assignment
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team: "network-ops"
+    object_id: 1
+    state: absent
+
+- name: Check if a team role assignment exists
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team: "network-ops"
+    object_id: 1
+    state: exists
+  register: result
+
+- name: Assign a role using numeric IDs
+  ansible.eda.role_team_assignment:
+    role_definition: "6"
+    team: "1"
+    object_id: 1
+    state: present
+
+- name: Assign a role using team_ansible_id
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team_ansible_id: "550e8400-e29b-41d4-a716-446655440000"
+    object_id: 1
+    state: present
+"""
+
+
+RETURN = r"""
+id:
+    description: ID of the role team assignment.
+    returned: when the assignment exists or was created
+    type: int
+    sample: 42
+role_definition:
+    description: The role definition ID.
+    returned: when state is 'exists' and the assignment is found
+    type: int
+    sample: 6
+team:
+    description: The team ID.
+    returned: when state is 'exists' and the assignment is found
+    type: int
+    sample: 1
+object_id:
+    description: The resource primary key the assignment is scoped to.
+    returned: when state is 'exists' and the assignment is found
+    type: str
+    sample: "1"
+content_type:
+    description: The content type of the scoped resource.
+    returned: when state is 'exists' and the assignment is found
+    type: str
+    sample: "shared.organization"
+object_ansible_id:
+    description: The Ansible resource UUID of the scoped resource.
+    returned: when state is 'exists' and the assignment is found
+    type: str
+    sample: null
+team_ansible_id:
+    description: The Ansible resource UUID of the team.
+    returned: when state is 'exists' and the assignment is found
+    type: str
+    sample: "550e8400-e29b-41d4-a716-446655440000"
+msg:
+    description: Message describing the result.
+    returned: when state is 'exists' and the assignment is not found
+    type: str
+    sample: "Role team assignment not found"
+"""
+
+
+from typing import Any, Optional
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ..module_utils.arguments import AUTH_ARGSPEC
+from ..module_utils.client import Client
+from ..module_utils.common import lookup_resource_id
+from ..module_utils.controller import Controller
+from ..module_utils.errors import EDAError
+
+
+def _resolve_id(
+    module: AnsibleModule,
+    controller: Controller,
+    endpoint: str,
+    value: str,
+    label: str,
+) -> Optional[int]:
+    """Resolve a name-or-ID string to an integer ID."""
+    try:
+        return int(value)
+    except (ValueError, TypeError):
+        pass
+
+    resource_id = lookup_resource_id(module, controller, endpoint, value)
+    if resource_id is None:
+        module.fail_json(msg=f"{label} '{value}' not found")
+    return resource_id
+
+
+def _find_existing(
+    controller: Controller,
+    filter_params: dict[str, Any],
+    object_id: Optional[int],
+    object_ansible_id: Optional[str] = None,
+) -> Optional[dict[str, Any]]:
+    """Find an existing role_team_assignment matching the filter and object."""
+    params = dict(filter_params, page_size=200)
+    response = controller.get_endpoint("role_team_assignments", data=params)
+
+    if response.status != 200:
+        return None
+
+    json_data = response.json
+    if not isinstance(json_data, dict):
+        return None
+
+    results: list[dict[str, Any]] = json_data.get("results", [])
+    for assignment in results:
+        if object_ansible_id is not None:
+            if str(assignment.get("object_ansible_id")) == str(object_ansible_id):
+                return assignment
+            continue
+
+        existing_obj_id = assignment.get("object_id")
+        if object_id is None and existing_obj_id is None:
+            return assignment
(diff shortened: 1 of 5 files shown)


plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,325 @@
...
+    if resource_id is None:
+        module.fail_json(msg=f"{label} '{value}' not found")
+    return resource_id
+
+
+def _find_existing(
+    controller: Controller,
+    filter_params: dict[str, Any],
+    object_id: Optional[int],
+) -> Optional[dict[str, Any]]:
```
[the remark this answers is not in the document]
B-Whitt: Addressed.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,325 @@
...
+
+    results: list[dict[str, Any]] = json_data.get("results", [])
+    for assignment in results:
+        existing_obj_id = assignment.get("object_id")
+        if object_id is None and existing_obj_id is None:
+            return assignment
+        if object_id is not None and str(existing_obj_id) == str(object_id):
+            return assignment
+
+    return None
```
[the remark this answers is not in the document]
B-Whitt: Addressed.

tests/integration/targets/role_team_assignment/tasks/main.yml
```diff
@@ -0,0 +1,173 @@
...
+    group/ansible.eda.eda:
+      aap_hostname: "{{ aap_hostname }}"
+      aap_username: "{{ aap_username }}"
+      aap_password: "{{ aap_password }}"
+      aap_validate_certs: "{{ aap_validate_certs }}"
+  block:
+    - name: Set test variables
+      ansible.builtin.set_fact:
+        test_team_name: "test-rta-team"
+        test_role_name: "Organization Auditor"
```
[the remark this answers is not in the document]
B-Whitt: Addressed.

tests/unit/plugins/modules/test_role_team_assignment.py
```diff
@@ -0,0 +1,466 @@
...
+                _make_api_list([]),
+            ],
+            post_response=Mock(status=201, json={"id": 77}),
+        )
+        _run_main(module, client)
+
+        post_data = client.post.call_args[1].get("data", {})
+        assert "object_id" not in post_data
+        assert post_data["object_ansible_id"] == obj_uuid
+
```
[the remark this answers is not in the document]
B-Whitt: Addressed.

tests/integration/targets/role_team_assignment/tasks/main.yml
```diff
@@ -0,0 +1,192 @@
...
+        status_code: [200, 403, 404, 405, 502]
+      register: gateway_check
+      ignore_errors: true
+
+    - name: Set teams API endpoint
+      ansible.builtin.set_fact:
+        teams_api_url: >-
+          {{ (gateway_check.status == 200) |
+             ternary(aap_hostname ~ '/api/gateway/v1/teams/',
+                     aap_hostname ~ '/api/eda/v1/teams/') }}
```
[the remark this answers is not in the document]
B-Whitt: Investigated against `aap-gateway` (`stable-2.7`), `django-ansible-base`, and the staging nginx config.

**403 can't happen on GET `/teams/`.** Both permission classes unconditionally allow safe methods — `IsSuperuserOrManageOrgsEnabled` ([`permissions.py:29-30`](https://github.com/ansible-automation-platform/aap-gateway/blob/stable-2.7/aap_gateway_api/utils/views/permissions.py#L29-L30)) checks `SAFE_METHODS` and `AnsibleBaseObjectPermissions` (`ansible_base/rbac/api/permissions.py:103-104`) returns `True` for all non-POST. Verified on live AAP 2.7: bad creds → 401, never 403.

**405 can't happen from Gateway on GET `/teams/`.** `TeamViewSet` inherits DRF `ModelViewSet` ([`common.py:19-33`](https://github.com/ansible-automation-platform/aap-gateway/blob/stable-2.7/aap_gateway_api/views/api/v1/common.py#L19-L33)) which always supports GET. The 405 in CI came from staging's `nginx`, not Gateway.

**However, status code alone isn't sufficient.** Staging's nginx has no `/api/gateway/` location — the catch-all serves `index.html` with status 200 and `content-type: text/html` ([`default.conf.template`](https://github.com/ansible/eda-server/blob/main/tools/docker/nginx/default.conf.template)). A status-only check would false-positive. Fixed by also checking `content_type` contains `application/json`:

```yaml
{{ (gateway_check.status == 200 and
    gateway_check.content_type is defined and
    'application/json' in gateway_check.content_type) | ... }}
```

Verified on live AAP 2.7 — all 21 test tasks pass.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,339 @@
...
+                payload["object_ansible_id"] = object_ansible_id
+
+            response = controller.post_endpoint(
+                "role_team_assignments",
+                data=payload,
+            )
+
+            if response.status in [200, 201]:
+                module.exit_json(changed=True, id=response.json["id"])
+            else:
```
B-Whitt: **Major** — Dead error-handling branch (inherited pattern)

`Client.post()` raises `EDAHTTPError` for non-success status codes before this `else` branch can execute. The custom error message here is unreachable dead code. Errors still surface via the `except EDAError` handler on line 334, but with a less descriptive message.

The same pattern exists in `credential_input_source.py`, so this is consistent with the collection. Non-blocking, but worth noting — the operator-friendly error message never reaches the user.

Same applies to the `else` branch after `delete_endpoint` on line 319.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,339 @@
...
+    team: "network-ops"
+    object_id: 1
+    state: present
+
+- name: Assign Organization Auditor using team name and org ID
+  ansible.eda.role_team_assignment:
+    role_definition: "Organization Auditor"
+    team: "network-ops"
+    object_id: 1
+    state: present
```
B-Whitt, line 94: **Minor** — Duplicate example

This example (lines 85-90) is functionally identical to the first example (lines 78-83) — both assign "Organization Auditor" to "network-ops" with `object_id: 1`, `state: present`. Consider replacing one with a different scenario, such as using `object_ansible_id` or omitting `object_id` for roles that don't require resource scoping.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+def _find_existing(
+    controller: Controller,
+    filter_params: dict[str, Any],
+    object_id: Optional[int],
+    object_ansible_id: Optional[str] = None,
+) -> Optional[dict[str, Any]]:
+    """Find an existing role_team_assignment matching the filter and object."""
+    params = dict(filter_params, page_size=200)
```
B-Whitt, line 208: **Nit** — Single-page lookup

`page_size=200` without following `next` links. Acceptable given no other module in the collection paginates, and 200+ assignments per role+team is an extreme edge case. The API would reject a duplicate POST with 400 anyway, so the failure mode is noisy rather than silent.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,339 @@
...
+            controller,
+            filter_params,
+            object_id,
+            object_ansible_id,
+        )
+
+        if state == "present":
+            if existing:
+                module.exit_json(changed=False, id=existing["id"])
+                return
```
B-Whitt, line 313: **Nit** — `return` after `module.exit_json()` is unreachable since `exit_json` calls `sys.exit()` internally. Harmless readability choice — same pattern at lines 277-278.

B-Whitt (commented):
## Code Review: feat(rbac): add role_team_assignment module for AAP 2.5

**Verdict**: READY_FOR_HUMAN_REVIEW
**Score**: 8/10 (Functionality 7 | Security 9 | Quality 8)

### Findings
- **1 Major** — Dead error-handling branches after `post_endpoint`/`delete_endpoint` (inherited from `credential_input_source.py` pattern). Non-blocking.
- **1 Minor** — Duplicate EXAMPLES entries
- **2 Nit** — Single-page lookup note, unreachable `return` statements

### Highlights
- Clean composite-key pattern, better factored than `credential_input_source.py`
- Proper mutual exclusivity for team/team_ansible_id and object_id/object_ansible_id
- UUID support correctly threaded through filters, matching, and payloads
- 22 unit tests exercising real function logic + integration tests with dynamic team names
- check_mode correctly handled for all three states

No critical or blocking issues. See inline comments for details.

---
_Reviewed in API-only mode (no local checkout). Codebase-wide searches and lint checks were not performed._

plugins/modules/role_team_assignment.py
B-Whitt: Addressed.

plugins/modules/role_team_assignment.py
B-Whitt: Addressed.

plugins/modules/role_team_assignment.py
B-Whitt: Skipping — accepted as non-blocking. No other module in the collection paginates, and the failure mode (duplicate POST → 400) is noisy rather than silent.

plugins/modules/role_team_assignment.py
B-Whitt: Skipping — harmless readability choice, consistent with the collection's existing pattern.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,343 @@
...
+                payload["object_id"] = object_id
+            elif object_ansible_id is not None:
+                payload["object_ansible_id"] = object_ansible_id
+
+            response = controller.post_endpoint(
+                "role_team_assignments",
+                data=payload,
+            )
+
+            if response.status in [201, 202, 204]:
```
wfealdel: **Minor**: The `else` branches after `post_endpoint` (lines 301-305) and `delete_endpoint` (lines 322-326) are unreachable. `Client.post()` and `Client.delete()` raise `EDAHTTPError` for non-success status codes before these branches execute — the `except EDAError` handler at line 338 catches the error instead, producing a less informative message (no HTTP status code or "Failed to create/delete" context). This is inherited from the `credential_input_source.py` pattern, so non-blocking, but worth noting.

Additionally, including 204 in the POST success check (`[201, 202, 204]`) would cause a `KeyError` on `response.json["id"]` since 204 responses have no body — though this is also unreachable for the same reason.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,343 @@
...
+    except (ValueError, TypeError):
+        pass
+
+    resource_id = lookup_resource_id(module, controller, endpoint, value)
+    if resource_id is None:
+        module.fail_json(msg=f"{label} '{value}' not found")
+    return resource_id
+
+
+def _find_existing(
```
wfealdel, line 201: **Minor**: `_find_existing` fetches a single page (`page_size=200`) without following `next` links. If there are >200 assignments for the same role_definition+team combination, a match could be missed, breaking idempotency. Extremely unlikely in practice and consistent with `credential_input_source.py`, but worth documenting the assumption.

plugins/modules/role_team_assignment.py
```diff
@@ -0,0 +1,343 @@
...
+    module: AnsibleModule,
+    controller: Controller,
+    endpoint: str,
+    value: str,
+    label: str,
+) -> Optional[int]:
+    """Resolve a name-or-ID string to an integer ID."""
+    try:
+        return int(value)
+    except (ValueError, TypeError):
```
wfealdel, line 192: **Minor**: `RETURN` only documents `id`, but `state=exists` returns additional fields (`role_definition`, `team`, `object_id`, `content_type`) via the `**details` unpack at line 331. The `credential_input_source` module documents its equivalent extra fields.

tests/unit/plugins/modules/test_role_team_assignment.py
```diff
@@ -0,0 +1,496 @@
...
+        "controller_token": None,
+        "request_timeout": 10.0,
+        "validate_certs": True,
+        "role_definition": "Organization Auditor",
+        "team": "test-team",
+        "team_ansible_id": None,
+        "object_id": 1,
+        "object_ansible_id": None,
+        "state": "present",
+        "update_secrets": True,
```
wfealdel: **Nit**: The mock params include `update_secrets: True`, which is not a parameter of this module. Harmless (MagicMock ignores it) but slightly misleading — consider removing it.

wfealdel (commented):
@B-Whitt

plugins/modules/role_team_assignment.py
B-Whitt: removed dead code

plugins/modules/role_team_assignment.py
B-Whitt: skipped. comments above.

plugins/modules/role_team_assignment.py
B-Whitt: additional fields documented

tests/unit/plugins/modules/test_role_team_assignment.py
B-Whitt: removed

AlexSCorey (approved):
This looks good to me.  I do worry about pagination in the  [here](https://github.com/ansible/event-driven-ansible/pull/554/changes#diff-d6df851a7ae580d30f74ff0480b127f814248b76ccb0eb209c533c43b4b0a1d0R208), but I think its an edge case that probably won't come up.

wfealdel (approved):

PR #556, opened 2026-06-26
fix(controller): detect embedded $encrypted$ tokens in field comparison

B-Whitt:
## Summary
- Fixes the root cause of flaky staging CI failures where project delete returns HTTP 409 "Project Sync is running" (AAP-75500)
- The `$encrypted$` detection in `has_encrypted_values()` and `fields_could_be_same()` only matched when the entire field value was `$encrypted$`, missing it when embedded in URLs (e.g., proxy field `http://user:$encrypted$@host:3128`)
- This caused spurious PATCH requests on idempotent module calls, which included unchanged `url` in the payload, triggering an unnecessary server-side project re-sync

## Root Cause
1. Project created with `proxy: "http://eda:testpass@squid:3128"`
2. API returns `proxy: "http://eda:$encrypted$@squid:3128"` (password masked)
3. Idempotent "create again" call compares masked vs unmasked proxy — `$encrypted$` substring not recognized
4. Module sends spurious PATCH with all fields including unchanged `url`
5. Server sees `url` in update_fields → triggers project re-sync → `import_state = PENDING`
6. Subsequent DELETE fails with HTTP 409

## Fix
- Skip fields in `objects_could_be_different()` where `$encrypted$` is embedded as a substring (not the whole value), preventing the false-positive diff
- Extend `has_encrypted_values()` and `fields_could_be_same()` to detect `$encrypted$` as a substring within string values
- Preserves existing `update_secrets=True` behavior for fully-encrypted fields (`$encrypted$` as the entire value)

## Verified
- **Live instance**: idempotent call with proxy now reports `changed: false` and `import_state` stays `completed`
- **CI log comparison**: confirmed main branch failures (PR #554, scheduled runs) show `import_state` regressing after idempotent calls; PR #542's wait loop absorbs the race but doesn't fix the root cause
- **Unit tests**: 262 passed, 0 failed, 0 regressions

## Test plan
- [x] Unit tests for `has_encrypted_values()` with embedded `$encrypted$`
- [x] Unit tests for `fields_could_be_same()` with embedded `$encrypted$`
- [x] Unit tests for `objects_could_be_different()` with proxy-like fields (both `update_secrets=True` and `False`)
- [x] Tox unit tests pass (262 passed)
- [x] Live instance validation against AAP 2.7
- [ ] Staging CI passes (pending merge)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Assisted-by: Claude Opus 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of encrypted values when the special marker appears inside a string, not just as the entire value.
  * Updated comparison behavior so records with embedded encrypted placeholders are handled more consistently.
  * Reduced false differences when comparing objects that include partially masked secret values.

* **Tests**
  * Expanded unit coverage for encrypted-value detection and object comparison across strings, lists, dictionaries, and mixed secret formats.

### DIFF
--- a/plugins/module_utils/controller.py
+++ b/plugins/module_utils/controller.py
@@ -224,6 +224,8 @@ class Controller:
                     return True
         elif obj == Controller.ENCRYPTED_STRING:
             return True
+        elif isinstance(obj, str) and Controller.ENCRYPTED_STRING in obj:
+            return True
         return False
 
     @staticmethod
@@ -242,6 +244,8 @@ class Controller:
             return True  # all sub-fields are either equal or could be equal
         if old_field == Controller.ENCRYPTED_STRING:
             return True
+        if isinstance(old_field, str) and Controller.ENCRYPTED_STRING in old_field:
+            return True
 
         return bool(new_field == old_field)
 
@@ -259,6 +263,12 @@ class Controller:
             old_field = old.get(field, None)
 
             if old_field != new_field:
+                if (
+                    isinstance(old_field, str)
+                    and Controller.ENCRYPTED_STRING in old_field
+                    and old_field != Controller.ENCRYPTED_STRING
+                ):
+                    continue
                 if self.update_secrets or (
                     not self.fields_could_be_same(old_field, new_field)
                 ):
--- a/tests/unit/plugins/module_utils/test_controller.py
+++ b/tests/unit/plugins/module_utils/test_controller.py
@@ -172,6 +172,26 @@ def test_has_encrypted_values() -> None:
     assert Controller.has_encrypted_values({"key": "value"}) is False
 
 
+@pytest.mark.parametrize(
+    "obj, expected",
+    [
+        ("$encrypted$", True),
+        ("http://user:$encrypted$@host:3128", True),
+        ("prefix$encrypted$suffix", True),
+        ("no_secrets_here", False),
+        ("", False),
+        (42, False),
+        (None, False),
+        (["$encrypted$"], True),
+        (["http://user:$encrypted$@host:3128"], True),
+        (["plain"], False),
+        ({"key": "http://user:$encrypted$@host:3128"}, True),
+    ],
+)
+def test_has_encrypted_values_substring(obj: object, expected: bool) -> None:
+    assert Controller.has_encrypted_values(obj) is expected
+
+
 def test_fail_wanted_one(mock_client: Mock, controller: Controller) -> None:
     response = MagicMock()
     response.json.return_value = {"count": 2, "results": [{"id": 1}, {"id": 2}]}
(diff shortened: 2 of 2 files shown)


plugins/module_utils/controller.py
```diff
@@ -259,6 +263,12 @@ def objects_could_be_different(
             old_field = old.get(field, None)
 
             if old_field != new_field:
+                if (
+                    isinstance(old_field, str)
+                    and Controller.ENCRYPTED_STRING in old_field
+                    and old_field != Controller.ENCRYPTED_STRING
+                ):
+                    continue
```
B-Whitt, line 271: **[Major] `update_secrets=True` bypass for embedded encrypted fields**

This `continue` block fires unconditionally — before the `self.update_secrets` check at the next `if` statement. When `update_secrets=True`, the operator explicitly wants to force-push secret values, but embedded-encrypted fields are silently skipped.

The exact-match `$encrypted$` path (excluded by the `old_field != Controller.ENCRYPTED_STRING` guard) correctly falls through to the existing `update_secrets` logic. This new embedded path does not.

```suggestion
                if (
                    isinstance(old_field, str)
                    and Controller.ENCRYPTED_STRING in old_field
                    and old_field != Controller.ENCRYPTED_STRING
                    and not self.update_secrets
                ):
                    continue
```

If this is intentional (embedded tokens can never be meaningfully compared), please document the rationale — the current test comment confirms the behavior but doesn't explain *why*.

plugins/module_utils/controller.py
```diff
@@ -224,6 +224,8 @@ def has_encrypted_values(obj: Any) -> bool:
                     return True
         elif obj == Controller.ENCRYPTED_STRING:
             return True
+        elif isinstance(obj, str) and Controller.ENCRYPTED_STRING in obj:
+            return True
```
B-Whitt, line 228: **[Minor] DRY: substring encrypted check repeated in 3 locations**

`isinstance(x, str) and Controller.ENCRYPTED_STRING in x` appears here, in `fields_could_be_same()` (line 247), and in `objects_could_be_different()` (line 267). Consider extracting a `_has_embedded_encrypted(value)` static method to reduce duplication and maintenance risk.

tests/unit/plugins/module_utils/test_controller.py
```diff
@@ -172,6 +172,26 @@ def test_has_encrypted_values() -> None:
...
+        ("$encrypted$", True),
+        ("http://user:$encrypted$@host:3128", True),
+        ("prefix$encrypted$suffix", True),
+        ("no_secrets_here", False),
+        ("", False),
+        (42, False),
+        (None, False),
+        (["$encrypted$"], True),
+        (["http://user:$encrypted$@host:3128"], True),
+        (["plain"], False),
```
B-Whitt, line 187: **[Nit] Missing edge case: list of dicts**

The parametrized test covers strings, lists, and dicts, but not `[{"key": "http://user:$encrypted$@host:3128"}]` (list of dicts with embedded encrypted values). The recursion should handle it, but it's a gap in the test matrix.

B-Whitt (commented):
**Code Review: 9.5/10 — NEEDS_CHANGES**

Functionality 9/10 | Security 10/10 | Quality 9.5/10

1 Major finding (embedded `$encrypted$` bypass of `update_secrets=True`), 1 Minor (DRY), 1 Nit (test gap). See inline comments for details.

*Generated with AI assistance: Claude Code / Opus 4.6 (Anthropic)*

plugins/module_utils/controller.py
ptoscano, line 271: Hm what is this check needed for? It seems to me that it is duplicating what `fields_could_be_same()` does, including the changes in this PR.

ptoscano (commented):
> Fixes the root cause of flaky staging CI failures where project delete returns HTTP 409 "Project Sync is running" (AAP-75500)

I don't understand what's the relation between that AAP ticket (which already has #542 as fix for it) with the changes of this PR. According to the explanation and the changes of this PR, the failures fixed should happen (almost) every time and with a different message, rather than being time-sensitive.

Would you please explicitly link/mention logs of failures or anything else that would suggest that `$encrypted$` in fields are responsible for this?

B-Whitt:
the problem seems to have gone away

